PR #13, opened 2018-11-08
Modify the "metrics-middleware" module to make the master version compile successfully.

jinlongchen:
Modify the "metrics-middleware" module to make the master version compile successfully.


aeneasr:
Sorry, what does this solve? This seems to be using an outdated version. Please use go11.1 with the ory ecosystem, and set `export GO111MODULE=on`.

PR #36, opened 2019-04-10
add pagination functions for use in ory/hydra

kminehart:
Signed-off-by: Kevin Minehart <kmineh0151@gmail.com>

## Related issue

https://github.com/ory/hydra/issues/1047

## Proposed changes

Add a function to create pagination a "Link" header given an offset and a limit.

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
vulnerability, I confirm that I got green light (please contact [hi@ory.sh](mailto:hi@ory.sh)) from the maintainers to push the changes.
- [x] I signed the [Developer's Certificate of Origin](https://github.com/ory/keto/blob/master/CONTRIBUTING.md#developers-certificate-of-origin)
by signing my commit(s). You can amend your signature to the most recent commit by using `git commit --amend -s`. If you
amend the commit, you might need to force push using `git push --force HEAD:<branch>`. Please be very careful when using
force push.

## Further comments

This is a necessary addition in order to complete the issue in Hydra. I figured this was a good place to add the header functionality, in case any other Ory products might use this same type of pagination.

If it should stay in Hydra that would be acceptable as well. Up to you!

I would verify that the logic in place in the tests matches with what you had in mind, @aeneasr

### DIFF
--- a/pagination/header.go
+++ b/pagination/header.go
@@ -20,13 +20,18 @@ func header(u *url.URL, rel string, limit, offset int) string {
 // If total is not set, then no "last" page will be calculated.
 // If no limit is provided, then this function will return an empty map.
 func Header(u *url.URL, total int, limit, offset int) http.Header {
-	if limit == 0 {
-		return http.Header{}
+	if limit <= 0 {
+		limit = 1
 	}
 
-	// lastOffset will either equal the offset required to contain the remainer,
+	// lastOffset will either equal the offset required to contain the remainder,
 	// or the limit.
-	lastOffset := total + (limit - total%limit) - limit
+	var lastOffset int
+	if total%limit == 0 {
+		lastOffset = total - limit
+	} else {
+		lastOffset = ((total / limit) * limit)
+	}
 
 	// Check for last page
 	if offset >= lastOffset {
--- a/pagination/header.go
+++ b/pagination/header.go
@@ -15,10 +15,38 @@ func header(u *url.URL, rel string, limit, offset int) string {
 	return fmt.Sprintf("<%s>; rel=\"%s\"", u.String(), rel)
 }
 
-func Header(u *url.URL, count int, limit, offset int) http.Header {
-	lastOffset := count + (limit - count%limit) - limit
+// Header returns an http header map, with a single key, "Link". The "Link" header is used in pagination where backwards compatibility is required.
+// The Link header will contain links any combination of the first, last, next, or previous (prev) pages in a paginated list (given a limit and an offset, and optionally a total).
+// If total is not set, then no "last" page will be calculated.
+// If no limit is provided, then this function will return an empty map.
+func Header(u *url.URL, total int, limit, offset int) http.Header {
+	if limit == 0 {
+		return http.Header{}
+	}
 
-	// Check for first page
+	// lastOffset will either equal the offset required to contain the remainer,
+	// or the limit.
+	lastOffset := total + (limit - total%limit) - limit
+
+	// Check for last page
+	if offset >= lastOffset {
+		if total == 0 {
+			return http.Header{
+				"Link": []string{
+					header(u, "first", limit, 0),
+					header(u, "next", limit, ((offset/limit)+1)*limit),
+					header(u, "prev", limit, ((offset/limit)-1)*limit),
+				},
+			}
+		}
+		return http.Header{
+			"Link": []string{
+				header(u, "first", limit, 0),
+				header(u, "prev", limit, lastOffset-limit),
+			},
+		}
+	}
+
 	if offset < limit {
 		return http.Header{
 			"Link": []string{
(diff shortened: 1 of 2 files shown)


pagination/header.go
```diff
@@ -0,0 +1,49 @@
...
+			"Link": []string{
+				header(u, "first", limit, 0),
+				header(u, "prev", limit, lastOffset-limit),
+			},
+		}
+	}
+
+	return http.Header{
+		"Link": []string{
+			header(u, "prev", limit, ((offset/limit)-1)*limit),
```
aeneasr: This will cause `+Inf` if limit is 0, maybe make sure limit can't be 0?

pagination/header.go
```diff
@@ -0,0 +1,49 @@
...
+
+func header(u *url.URL, rel string, limit, offset int) string {
+	q := u.Query()
+	q.Set("limit", strconv.Itoa(limit))
+	q.Set("offset", strconv.Itoa(offset))
+	u.RawQuery = q.Encode()
+	return fmt.Sprintf("<%s>; rel=\"%s\"", u.String(), rel)
+}
+
+func Header(u *url.URL, count int, limit, offset int) http.Header {
```
aeneasr: Maybe rename count to total? What would happen if `count` is zero (e.g. because we did not count all objects in the database)?

aeneasr: `Header` should be documented: `// Header sets ...`

aeneasr: I think it would make sense to change the signature a bit:

```go
func SetHeader(header http.Header, count int, limit, offset int)
```

pagination/header_test.go
```diff
@@ -0,0 +1,76 @@
...
+		t.Fatal(err)
+	}
+
+	t.Run("Create previous and first but not next or last if at the end", func(t *testing.T) {
+		h := Header(u, 120, 50, 100)
+
+		expect := http.Header{
+			"Link": []string{
+				"<http://example.com?limit=50&offset=0>; rel=\"first\"",
+				"<http://example.com?limit=50&offset=50>; rel=\"prev\"",
```
aeneasr: I think the last element should not have a `,` appended

pagination/header.go
kminehart: I think we could either return an error or just not set `last`. We should probably choose to not set `last` as counting the total number of rows can be a pretty costly operation.

kminehart: In order to build the links I need the URL that was requested.

The usage I envisioned kind of went like this:

```
func handler(r *http.Request, w http.ResponseWriter) {
    limit, offset := pagination.Parse(r, 100, 0, 500)

    // n = total
    c, n, err := h.r.ClientManager().GetClients(r.Context(), limit, offset)

    // ...

    pagination.Header(r.URL, n, limit, offset).Write(w)
    h.r.Writer().Write(w, r, c)
}
```

pagination/header_test.go
kminehart: are you referring to the comma delimiting the go slice?

pagination/header.go
kminehart: ah, right. I'll fix that.

pagination/header_test.go
aeneasr: My bad 😅

pagination/header.go
aeneasr: Oh I didn't know that works! That's great!

pagination/header.go
```diff
@@ -15,35 +15,54 @@ func header(u *url.URL, rel string, limit, offset int) string {
...
 }
 
-func Header(u *url.URL, count int, limit, offset int) http.Header {
-	lastOffset := count + (limit - count%limit) - limit
+// Header returns an http header map, with a single key, "Link". The "Link" header is used in pagination where backwards compatibility is required.
+// The Link header will contain links any combination of the first, last, next, or previous (prev) pages in a paginated list (given a limit and an offset, and optionally a total).
+// If total is not set, then no "last" page will be calculated.
+// If no limit is provided, then this function will return an empty map.
+func Header(u *url.URL, total int, limit, offset int) http.Header {
+	if limit == 0 {
```
aeneasr: Maybe just show prev/next if limit is 0?

pagination/header_test.go
kminehart: :joy:

pagination/header.go
kminehart: I think we could set a default limit (10?) or we could use the offset as the default limit.

what do you think?

aeneasr: Sorry, I missread, I'd simply put limit to 1 if the value is 0 or negative!

aeneasr: Alternatively only provide `first`?

kminehart: I like the idea of setting it to 1.

pagination/header.go
kminehart: I've renamed count to total, and if total is 0, then no `last` page is provided.

aeneasr:
Awesome!

PR #49, opened 2019-05-05
Prometheus metrics

kminehart:
## Related issue

https://github.com/ory/hydra/issues/1176

## Proposed changes

https://github.com/ory/x/pull/49#issuecomment-493837126

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
vulnerability, I confirm that I got green light (please contact [hi@ory.sh](mailto:hi@ory.sh)) from the maintainers to push the changes.
- [x] I signed the [Developer's Certificate of Origin](https://github.com/ory/keto/blob/master/CONTRIBUTING.md#developers-certificate-of-origin)
by signing my commit(s). You can amend your signature to the most recent commit by using `git commit --amend -s`. If you
amend the commit, you might need to force push using `git push --force HEAD:<branch>`. Please be very careful when using
force push.
    - Worth noting that a lot of this would be difficult to usefully test; it can be done if needed, though!


metricsx/bridge_segment.go
```diff
@@ -0,0 +1,238 @@
...
+)
+
+type analyticsProperties struct {
+	memstats analytics.Properties
+	app      analytics.Properties
+}
+
+// This map is for converting Prometheus metrics (by name) to segment keys that were
+// previously sent from ory/x/metricsx
+var memStats = map[string]string{
```
aeneasr: Reading concurrently from maps can cause panics as it is not specified by the language definition that maps are thread safe at all.

metricsx/bridge_segment.go
```diff
@@ -0,0 +1,238 @@
...
+
+func getSegmentProperties(mfs []*dto.MetricFamily) analyticsProperties {
+	p := analyticsProperties{
+		memstats: analytics.NewProperties(),
+		app:      analytics.NewProperties(),
+	}
+	for _, v := range mfs {
+		for _, m := range v.GetMetric() {
+			if v.GetType() == dto.MetricType_HISTOGRAM || v.GetType() == dto.MetricType_SUMMARY {
+				// Don't transfer histogram buckets / summary percentiles to segment
```
aeneasr: Is there a way to ensure that we only send stuff to segment that segment actually understands? Are there more types than the listed ones maybe?

kminehart: I've come up with a much better solution to this. :)

kminehart:
@aeneasr I believe it's at a good stopping point now.

My approach here was intended to solve a few problems:

1. Sending lots of data to segment is expensive.
2. Some Prometheus metrics should be tracked with Segment, while others may not.
3. These changes should not affect other applications that import this file (for now). I'm sure a cleanup effort here could be made, but in the meantime, maintaining the same signatures is important.
4. Flexibility to offer solutions other than Prometheus, while treating Prometheus metrics as a first-class citizen.

Basically, I wrapped the 2 main prometheus types and its variations (Gauge and Counter). For Counters and Gauges intended to be tracked, replace `prometheus.NewGauge...(...)` and `prometheus.NewCounter...(...)` calls with the same signature, only using `metricsx` instead.

Using data sent previously, these functions will add wrappers around the Gauge and Counter type. The only one supported currently is Segment for hydra telemetry.

Another thing that was added was the `Observer` type and `Watch` function. These two will basically update metrics that poll the database at an interval. This solves the problem of some metrics only being collected when the `/metrics/prometheus` page was hit.

aeneasr:
Thank you! I didn't have time to review this yet due to travel but will do so until the weekend :)

metricsx/middleware.go
```diff
@@ -71,256 +44,90 @@ func Hash(value string) string {
...
-
-	next(rw, r)
-
-	if sw.optOut {
-		return
-	}
-
-	latency := time.Now().UTC().Sub(start.UTC()) / time.Millisecond
-
+func enqueue(
```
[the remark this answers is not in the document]
kminehart: Yeah! Originally I decoupled it because I used the middleware separately / reused it. Since I've changed my approach, this stayed the same. I'll change it back to how it was.

kminehart:
Did some research, I think what I have is sufficient for keeping a thread-safe implementation.

Multiple threads shouldn't have any issue calling `client.Enqueue`. The one write to that pointer is behind a mutex.

I really wanted to keep all of the client.Enqueue calls away from a mutex; I think that would have a really performance impact.

Thanks, @aeneasr!

PR #52, opened 2019-05-09
migratest: add FindMatchingTestMigrations func

lopezator:
A few comments here:

- `FindMatchingTestMigrations` gets a PackrMigrationSource as input, for example:

* migrations/sql/10.sql, migrations/sql/11.sql, migrations/sql/12.sql

And returns its matching test files instead of the full folder:

* migrations/sql/tests/10_test.sql, migrations/sql/tests/11_test.sql, migrations/sql/tests/12_test.sql

CC\ @aeneasr


lopezator:
Just a couple of notes more:

If we don't want to hide here the logic that the test file names must be named as 1_test.sql, 2_test.sql, etc...

A. We could pass the replace pattern as a parameter somehow to the func.
B. If we drop the "_test" from test files, and renane them all as:

1.sql, instead of 1_test.sql

I think we could avoid having to convert the filenames here.

aeneasr:
That's totally true but I'm also super happy with the implementation as is! The only thing I'd like to add would be a test to make sure everything works as intended. I kind of like the `_test` pattern because it makes it very explicit that this is a test and it also follows go convention :)

lopezator:
@aeneasr tests added, let me know if you have more suggestions!

aeneasr:
🙌 🙌 🙌

PR #57, opened 2019-06-05
Fix the DB username can't contain "@" character

yxzhm:
If  user name of DB contains"@" char, the sqlcon can't connect to the DB.
The DSN example is as following:
DSN=mysql://sa@test.mysql.database.azure.com:MyPassword@(127.0.0.1:3306)/hydra

This PR do a very simple fix.

### DIFF
--- a/sqlcon/connector.go
+++ b/sqlcon/connector.go
@@ -196,7 +196,7 @@ func connectionString(clean *url.URL) string {
 	userinfo := username
 	password, hasPassword := clean.User.Password()
 	if hasPassword {
-		userinfo = url.QueryEscape(userinfo) + ":" + url.QueryEscape(password)
+		userinfo = userinfo + ":" + password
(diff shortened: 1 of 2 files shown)


sqlcon/connector.go
```diff
@@ -196,7 +196,7 @@ func connectionString(clean *url.URL) string {
 	userinfo := username
 	password, hasPassword := clean.User.Password()
 	if hasPassword {
-		userinfo = url.QueryEscape(userinfo) + ":" + url.QueryEscape(password)
+		userinfo = userinfo + ":" + password
```
aeneasr: This should only happen when the scheme is `mysql`

yxzhm: Done the reword, based on your comments

aeneasr:
Thank you!

PR #85, opened 2019-10-26
viperx: Added viper BindEnvsToSchema helper

aeneasr:
See https://github.com/ory/oathkeeper/issues/280
See https://github.com/ory/oathkeeper/issues/279
See https://github.com/ory/oathkeeper/issues/276
See https://github.com/ory/oathkeeper/issues/270


viperx/bind_env.go
```diff
@@ -0,0 +1,115 @@
...
+	"oneOf",
+}
+
+// BindEnvsToSchema uses all keys it can find from ``
+func BindEnvsToSchema(schema json.RawMessage) error {
+	keys, err := getSchemaKeys(string(schema), string(schema), []string{}, []string{})
+	if err != nil {
+		return err
+	}
+	return viper.BindEnv(keys...)
```
hypnoglow: As I noted previously this usage of BindEnv is incorrect: this function does not accept multiple keys https://github.com/spf13/viper/blob/master/viper.go#L960 It either accepts one parameter (the key) or two parameters (the key and the corresponding ENV_VAR_NAME)

hypnoglow: Also, how do we deal with the env names? I guess with the effect of `SetEnvKeyReplacer` it will work automatically, so after`viper.BindEnv("foo.bar.baz")` env `FOO_BAR_BAZ` should work, right?

aeneasr: Ah, thanks, I overlooked that completely. Fixed and added a test to make sure this works properly.

> Also, how do we deal with the env names? I guess with the effect of SetEnvKeyReplacer it will work automatically, so afterviper.BindEnv("foo.bar.baz") env FOO_BAR_BAZ should work, right?

Exactly!

PR #100, opened 2019-12-12
add helper method to get config with env vars

pike1212:
## Related issue

https://github.com/ory/oathkeeper/issues/305

## Proposed changes

Helper method that uses viper.AllSettings which will do env var replace since viper.GetStringMap does not.

## Checklist

vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.

### DIFF
--- a/viperx/helper_test.go
+++ b/viperx/helper_test.go
@@ -393,7 +393,7 @@ func TestGetStringMapConfig(t *testing.T) {
 					})
 				} else {
 					InitializeConfig("project-stub-name", "", l)
-					config := GetStringMapConfig("authenticators", "oauth2_introspection")
+					config := GetStringMapConfig("authenticators", "oauth2_introspection", "config")
 
 					assert.Equal(t, "http://myurl", config["introspection_url"])
 				}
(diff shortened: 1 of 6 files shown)


viperx/helper_test.go
```diff
@@ -365,3 +365,69 @@ func TestGetDuration(t *testing.T) {
...
+			{f: "./stub/yaml/.project-stub-name.yaml"},
+			{f: "./stub/yml/.project-stub-name.yml"},
+			{f: "./stub/does-not-exist/foo.yml", fatals: true},
+		} {
+			t.Run(fmt.Sprintf("case=%d/path=%s", k, tc.f), func(t *testing.T) {
+				viper.Reset()
+
+				cfgFile = tc.f
+				if tc.fatals {
+					assert.Panics(t, func() {
```
aeneasr: A fatal is unfortunately not a panic - I'm not sure if we can actually catch a fatal easily here. This causes the test to fail atm :/

pike1212: Issue was that cfgFile was being set, and tests further down the chain were runnign with it in a bad state, so I set it back to empty after the test

aeneasr:
Thank you! I've released this as v0.0.87 :)

PR #106, opened 2020-01-22
jsonschemax: add format and pattern to path

zepatrik:
## Related issue

https://github.com/ory/kratos/issues/175

## Proposed changes

The returned Path struct now includes the format and pattern defined by the schema


aeneasr:
Released as v0.0.90

PR #120, opened 2020-03-12
fix(viperx): add type conversion to make sure viper knows what type a value should be

zepatrik:
## Related issue
closes https://github.com/ory/viper/issues/13

The type conversion is very ugly and only works for string slices, but that's most of the cases.
I think it would be an awesome feature in viper to set a type regardless of the default value. That would make this conversion obsolete.

### DIFF
--- a/viperx/bind_env.go
+++ b/viperx/bind_env.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"reflect"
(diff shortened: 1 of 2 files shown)


zepatrik:
@ArthurKnoep

zepatrik:
ok @aeneasr we will probably have to make that change in viper

aeneasr:
Yeah let’s fix it in ory/viper!

PR #129, opened 2020-03-24
new: add support to sqlcon to handle postgreSQL serialization_failure error

aaslamin:
## Related issue

- https://github.com/ory/hydra/issues/1719
- https://github.com/ory/hydra/issues/1735 

## Proposed changes

Add new error handling for the serialization error returned by postgreSQL when using the `REPEATABLE READ` isolation level. This error is returned when an UPDATE/DELETE (TXN2) statement was queued behind another UPDATE/DELETE (TXN1) statement on the same row. Once TXN1 completes and commits, TXN2 will fail to ensure consistency.

- postgreSQL error code appendix: https://www.postgresql.org/docs/9.6/errcodes-appendix.html
- error code: 40001

## Dependent PR(s) ⚠️ 

This error will be consumed by https://github.com/ory/hydra/pull/1766 so it can then return a `fosite. ErrSerializationFailure` which will then be [consumed by `fosite` to render an appropriate RFC 6749 error](https://github.com/ory/fosite/pull/402).

## Checklist

- [ ] I have read the [security policy](../security/policy)
vulnerability, I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/sqlcon/error.go
+++ b/sqlcon/error.go
@@ -7,11 +7,9 @@ import (
 
 	"github.com/go-sql-driver/mysql"
 	"github.com/jackc/pgconn"
-	"github.com/pkg/errors"
-
 	"github.com/ory/herodot"
-
 	"github.com/ory/x/errorsx"
+	"github.com/pkg/errors"
 )
 
 var (
@@ -26,6 +24,12 @@ var (
 		CodeField:   http.StatusNotFound,
 		StatusField: http.StatusText(http.StatusNotFound),
 		ErrorField:  "Unable to locate the resource",
+	}
+	// ErrConcurrentUpdate is returned when the database is unable to serialize access due to a concurrent update.
+	ErrConcurrentUpdate = &herodot.DefaultError{
+		CodeField:   http.StatusBadRequest,
+		StatusField: http.StatusText(http.StatusBadRequest),
+		ErrorField:  "Unable to serialize access due to a concurrent update in another session",
 	}
 )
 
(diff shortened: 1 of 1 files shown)


aaslamin:
@aeneasr I noticed there is no test coverage here, I could add some if you like? ✨

aaslamin:
How do I get the Semantic PR build check to trigger again? I modified my commit message to follow the style and force pushed another commit, but no luck 😢

sqlcon/error.go
```diff
@@ -43,6 +47,8 @@ func HandleError(err error) error {
 		switch err.Code {
 		case "23505": // "unique_violation"
 			return errors.Wrap(ErrUniqueViolation, err.Error())
+		case "40001": // "serialization_failure"
```
aeneasr: Do we only care about PostgreSQL for now?

aeneasr:
Released as v0.0.108

aaslamin:
> LGTM! Should we add MySQL in another PR?

Yes. And also for cockroachDB if it is returning an error type different from what the postgreSQL driver is returning.

> Regarding tests: I'm not sure how to test the HandleError except for mocking the error which isn't very useful IMO. What do you think?

Yeah, that's what I was thinking. Value add is just to catch regressions due to things like accidental refactoring, etc.

> In any case, I'll merge this so you can continue working on the other PR!

Thank you! 🙌

sqlcon/error.go
aaslamin: See https://github.com/ory/x/pull/129#issuecomment-603427081

PR #142, opened 2020-04-26
feat(sqlconn): enhance ParseConnectionOptions to return cleaned DSN

zepatrik:
ParseConnectionOptions will also return the DSN URI with removed connection options.


aeneasr:
LGTM

PR #148, opened 2020-05-13
feat: added kratos' db json scanner functions

zepatrik:
also a `RawJSON` type is added


aeneasr:
Can we use this in Kratos then and remove the originals?

zepatrik:
sure that's the idea

PR #153, opened 2020-05-26
feat: add NullUUID type

zepatrik:
This new type wraps uuid.UUID to allow nullable UUID columns.


zepatrik:
Stupid me... this type is already implemented in `gofrs/uuid`

PR #173, opened 2020-07-03
fix: removing whitespaces in config file path

tricky42:
## Proposed changes
This fixes simply trims whitespaces from the configuration file path if one has been specified -c // --config flag. 
This fixes a problem if you try to run e.g. Kratos and specify a custom configuration will with -c // --config flag without specifying each part of the commandline call as a seperate string literal: 

`kratos serve "-c ./myconfig.yml"`

fails with the following error message, although the file exists and the process has sufficient permissions to read the file:

`Unable to open config file. Make sure it exists and the process has sufficient permissions to read it  audience=application error=map[message:open  ./myconfig.yml: no such file or directory] path= ./contrib/quickstart/kratos/email-password/.kratos.yml service_name=kratos service_version=`

This is caused by the handling of flags as defined by https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html, which results in a whitespace in the beginning of the config file path. 

As the error message is misleading and it is easy to overlook the whitespace in the error message, the fix simply removes whitespaces from the file path before it gets passed to viperx.

## Checklist
      the changes.
- [ ] I have added necessary documentation within the code base (if appropriate)

### DIFF
--- a/viperx/stub/yml/.project-stub-name-invalid-path.yml
+++ b/viperx/stub/yml/.project-stub-name-invalid-path.yml
@@ -1,0 +1,7 @@
+serve:
+  admin:
+    port: 1
+authenticators:
+  oauth2_introspection:
+    config:
+      introspection_url: http://myurl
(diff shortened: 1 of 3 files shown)


aeneasr:
Perfect PR, thank you!

aeneasr:
released as v0.0.137

PR #174, opened 2020-07-22
feat: debug output all viper settings

zepatrik:
This feature contributes to our better-debug-output-offensive. It is especially usefull for users to note that a value is not actually set or set in a wrong way.
Also on config changes the whole config will now be printed making it possible to find all differences that occurred.

### DIFF
--- a/viperx/validate.go
+++ b/viperx/validate.go
@@ -3,8 +3,9 @@ package viperx
 import (
 	"bytes"
 	"encoding/json"
-	"github.com/ory/x/logrusx"
 	"io/ioutil"
+
+	"github.com/ory/x/logrusx"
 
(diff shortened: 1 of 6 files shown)


viperx/validate.go
```diff
@@ -52,6 +54,7 @@ func Validate(name string, content []byte) error {
 	if err := json.NewEncoder(&b).Encode(viper.AllSettings()); err != nil {
 		return errors.WithStack(err)
 	}
+	l.WithFields(viper.AllSettings()).Debug("detected config values")
```
aeneasr, line 57: This is a pretty big breaking change for one debug statement, does that really make sense?

zepatrik:
reworked in #175

PR #175, opened 2020-07-23
feat: add debug config dumping to viperx

zepatrik:
## Related issue

This feature contributes to our better-debug-output-offensive. It is especially useful for users to note that a value is not actually set or set in a wrong way.
Also on config changes the whole config will be dumped making it possible to trace down differences.

## Proposed changes

WatchAndValidateViper will dump the whole config including sensitive values in a provided directory. Files explicitly have RW permission only for the owner.


viperx/output.go
```diff
@@ -76,3 +81,22 @@ func jsonschemaFormatError(e *jsonschema.ValidationError) (string, string) {
...
+func getDumpFileName(i int) string {
+	return fmt.Sprintf("config-%08d.yml", i)
+}
+
+func sensitiveDumpAllValues(dir string) error {
+	configContent, err := yaml.Marshal(viper.AllSettings())
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	lastDumpedConfig++
```
aeneasr, line 98: This is racy because it may be called by several goroutines simoultaneously. Use `atomic` if you want a global counter: https://golang.org/pkg/sync/atomic/ 

Why not just use the unix nano timestamp? `time.Now().UnixNano()` which is most likely unique enough?

viperx/viper.go
```diff
@@ -74,7 +74,7 @@ func WatchConfig(l *logrusx.Logger, o *WatchOptions) {
 		watcherLock.Lock()
 		defer watcherLock.Unlock()
 
-		l.WithField("file", in.Name).
+		defer l.WithField("file", in.Name).
 			WithField("operator", in.Op.String()).
 			WithField("immutables", o.Immutables).
 			Info("The configuration has changed and was reloaded.")
```
aeneasr, line 80: If you use defer, then `was reloaded` is misleading as it will appear on error as well. So let's change it to:

```suggestion
			Info("A change to the configuration file was processed.")
```

go.mod
```diff
@@ -4,6 +4,7 @@ require (
 	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/fatih/structs v1.1.0
 	github.com/fsnotify/fsnotify v1.4.9
+	github.com/ghodss/yaml v1.0.0
```
aeneasr, line 7: https://github.com/go-yaml/yaml is much more popular and also used by viper

zepatrik: right I used the same as in kratos

PR #176, opened 2020-07-28
feat: add watcherx/FileWatcher

zepatrik:
## Related issue

The FileWatcher is a kubernetes config map and macos tmp compatible file watcher.
Further we will add more watchers to watch websockets/...

### DIFF
--- a/watcherx/file.go
+++ b/watcherx/file.go
@@ -21,46 +21,46 @@ func WatchFile(ctx context.Context, file string, c EventChannel) error {
 	if err := w.Add(dir); err != nil {
 		return errors.WithStack(err)
 	}
-	realFile, err := filepath.EvalSymlinks(file)
+	resolvedFile, err := filepath.EvalSymlinks(file)
 	if err != nil {
-		if _, ok := err.(*os.PathError); ok {
-			// the file watcher should still watch the directory to get notified for
-			// file creation
-			realFile = ""
-		} else {
+		if _, ok := err.(*os.PathError); !ok {
 			return errors.WithStack(err)
 		}
-	}
-	if realFile != "" && realFile != file {
-		// file is a symlink, we have to explicitly watch the referenced file.
-		// We are watching file instead of lastFile because fsnotify identifies
-		// watch entries by the passed name but follows symlinks when watching
-		// (at least on unix but not on windows).
+		// The file does not exist. The watcher should still watch the directory
+		// to get notified about file creation.
+		resolvedFile = ""
+	} else if resolvedFile != file {
+		// If `resolvedFile` != `file` then `file` is a symlink and we have to explicitly watch the referenced file.
+		// This is because fsnotify follows symlinks and watches the destination file, not the symlink
+		// itself. That is at least the case for unix systems. See: https://github.com/fsnotify/fsnotify/issues/199
 		if err := w.Add(file); err != nil {
 			return errors.WithStack(err)
 		}
 	}
-	go streamFileEvents(ctx, w, c, file, realFile)
+	go streamFileEvents(ctx, w, c, file, resolvedFile)
 	return nil
 }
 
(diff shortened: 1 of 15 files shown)


watcherx/file.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	realFile, err := filepath.EvalSymlinks(file)
+	if err != nil {
+		if _, ok := err.(*os.PathError); ok {
+			// the file watcher should still watch the directory to get notified for
+			// file creation
+			realFile = ""
+		} else {
+			return errors.WithStack(err)
+		}
```
aeneasr, line 32: Best-practice Go code follows a pattern of "early return": https://dave.cheney.net/practical-go/presentations/qcon-china.html#_return_early_rather_than_nesting_deeply

```suggestion
			realFile = ""
		}
		return errors.WithStack(err)
```

watcherx/file.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if realFile != "" && realFile != file {
+		// file is a symlink, we have to explicitly watch the referenced file.
+		// We are watching file instead of lastFile because fsnotify identifies
+		// watch entries by the passed name but follows symlinks when watching
+		// (at least on unix but not on windows).
```
aeneasr, line 38: I think this needs clarification (please read and correct any misunderstanding):

```suggestion
		// If `realFile` != `file` then `file` is a symlink and we have to explicitly watch the referenced file.
		// This is because fsnotify follows symlinks and watches the destination file, not the symlink
		// itself. That is at least the case for unix systems. See: https://github.com/fsnotify/fsnotify/issues/199
```

watcherx/file.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			_ = watcher.Close()
+			return
+		case e, ok := <-watcher.Events:
+			if !ok {
+				_ = watcher.Close()
+				return
+			}
```
aeneasr, line 72: Move `path.Clean(e.Name) == file` check to here:

```
if path.Clean(e.Name) == file {
  goto loopEnd
}
```

watcherx/file.go
```diff
@@ -0,0 +1,118 @@
...
+		// (at least on unix but not on windows).
+		if err := w.Add(file); err != nil {
+			return errors.WithStack(err)
+		}
+	}
+	go streamFileEvents(ctx, w, c, file, realFile)
+	return nil
+}
+
+func streamFileEvents(ctx context.Context, watcher *fsnotify.Watcher, c EventChannel, file, lastFile string) {
```
aeneasr, line 47: Rename to:

```suggestion
// streamFileEvents watches for file changes and supports symlinks which requires several workarounds due to limitations of fsnotify.
// Argument `resolvedSymlinkPath` is the resolved symlink path of the file, or it is the fileWatched name itself. If `resolvedSymlinkPath` is empty, then the fileWatched, or the symlink pointing to the destination file do not exist. 
func streamFileEvents(ctx context.Context, watcher *fsnotify.Watcher, c EventChannel, fileWatched, resolvedSymlinkPath string) {
```

watcherx/file.go
```diff
@@ -0,0 +1,118 @@
...
+				_ = watcher.Close()
+				return
+			}
+			// e.Name contains the name we started the watcher with, not the actual file name
+			if path.Clean(e.Name) == file && e.Op&fsnotify.Remove != 0 {
+				// the file (or the file behind the symlink) was removed
+				c <- &RemoveEvent{eventSource}
+				goto loopEnd
+			}
+			currentFile, err := filepath.EvalSymlinks(file)
```
aeneasr, line 79:
```suggestion
			var err error
			recentlyResolvedSymlinkPath, err = filepath.EvalSymlinks(file)
```

PR #180, opened 2020-08-03
feat: add watcherx/DirectoryWatcher

zepatrik:
## Proposed changes

This new directory watcher notifies about all file changes in one directory up to a specified depth. It ignores events related to directory creation/deletion.

Current limitations:
- the directory has to exist for the watcher to start, it does not automatically watch the parent directory for a create event (though we can add that)
- for the same reason deleting the watched directory will make the watcher close itself
- it does not use the symlink compatible FileWatcher (though it probably should)

### DIFF
--- a/watcherx/directory.go
+++ b/watcherx/directory.go
@@ -4,92 +4,117 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/fsnotify/fsnotify"
-	"github.com/pkg/errors"
 	"io/ioutil"
 	"os"
-	"path/filepath"
+	"path"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/pkg/errors"
 )
 
-func listSubDirsDepth(dirs []string, errs []error, parent string, depth int) ([]string, []error) {
+func listSubDirsDepth(parent string, depth int) (dirs []string, errs []error) {
 	if depth == 0 {
-		return dirs, nil
+		return
 	}
-	files, err := ioutil.ReadDir(parent)
-
+	entries, err := ioutil.ReadDir(parent)
+	if err != nil {
+		return dirs, append(errs, errors.WithStack(err))
+	}
+	for _, e := range entries {
+		if e.IsDir() {
+			dn := path.Join(parent, e.Name())
+			subDirs, subErrs := listSubDirsDepth(dn, depth-1)
+			dirs = append(append(dirs, dn), subDirs...)
+			errs = append(errs, subErrs...)
+		}
+	}
+	return
 }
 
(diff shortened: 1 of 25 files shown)


watcherx/directory.go
```diff
@@ -0,0 +1,120 @@
...
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	subDirs, errs := listSubDirsDepth(dir, depth)
+	if len(errs) != 0 {
+		return errors.Errorf("%+v", errs)
+	}
+	for _, d := range append(subDirs, dir) {
+		fmt.Printf("adding watcher for %s\n", d)
```
aeneasr, line 45: Remove debug statement

watcherx/directory.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/pkg/errors"
+)
+
+func listSubDirsDepth(parent string, depth int) (dirs []string, errs []error) {
+	if depth == 0 {
+		return
+	}
+	entries, err := ioutil.ReadDir(parent)
+	if err != nil {
+		return dirs, append(errs, errors.WithStack(err))
```
aeneasr, line 22: Why not just return on an error immediately, given that you return directly here as well: https://github.com/ory/x/pull/180/files#diff-61f10616b3d8e8d2b3d31a972c41c59eR41

watcherx/directory.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/pkg/errors"
+)
+
+func listSubDirsDepth(parent string, depth int) (dirs []string, errs []error) {
+	if depth == 0 {
+		return
+	}
+	entries, err := ioutil.ReadDir(parent)
```
aeneasr, line 20: Have you seen `filepath.Walk`?

https://yourbasic.org/golang/list-files-in-directory/

watcherx/directory.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	} else if e.Op&(fsnotify.Write|fsnotify.Create) != 0 {
+		if stats, err := os.Stat(e.Name); err != nil {
+			c <- &ErrorEvent{
+				error:  errors.WithStack(err),
+				source: source(e.Name),
+			}
+			return
+		} else if stats.IsDir() {
+			fmt.Printf("created directory %s\n", e.Name)
```
aeneasr, line 86: Remove debug statement

watcherx/directory.go
zepatrik: Hm right for some reason I decided to not use it, I think because I misread the documentation and thought it would only list files and not directories.

PR #190, opened 2020-08-17
feat: add websocket watcher

zepatrik:
The websocket watcher is just any of the other watchers but piped through a websocket.


watcherx/websocket_client.go
```diff
@@ -0,0 +1,51 @@
...
+
+func closeOnDone(ctx context.Context, conn *websocket.Conn) {
+	<-ctx.Done()
+	conn.Close()
+}
+
+func forwardWebsocketEvents(conn *websocket.Conn, c EventChannel, u *url.URL) {
+	serverURL := source(u.String())
+	defer conn.Close()
+	for {
```
aeneasr, line 29: This queue never finishes. Maybe listen on `context.Done()`?

zepatrik: That's why there is https://github.com/ory/x/pull/190/files#diff-5db460a2bf191856555ebe1ceb20c2b3R16

See https://github.com/gorilla/websocket/issues/71 as a reference.

zepatrik: `conn.ReadMessage` is blocking

aeneasr: So `WatchWebsocket` will just accumulate go routines that can not be killed? Not sure if I like that :D

aeneasr:
fixed on master you can rebase

zepatrik:
Done

aeneasr (approved):

PR #219, opened 2020-10-22
fix: Functions to parse URLs and handle URL file paths

PatrLind:
This code was first created for PR #557 in oathkeeper.
It should probably be in the shared ory/x project instead of there,
so that is why this code is checked in here.

## Related PR: https://github.com/ory/oathkeeper/pull/557

Related Ory person: @zepatrik 

      the changes.
- [ ] I have added necessary documentation within the code base (if appropriate)

## Further comments

See other PR for more details

### DIFF
--- a/urlx/parse.go
+++ b/urlx/parse.go
@@ -25,12 +25,10 @@ func Parse(rawURL string) (*url.URL, error) {
 		return url.Parse(rawURL)
 	}
 
-	if strings.HasPrefix(lcRawURL, "file://") {
-		// Normally the first part after file:// is a hostname, but since
-		// this is often misused we interpret the URL like a normal path
-		// by removing the "file://" from the beginning
-		rawURL = rawURL[7:]
-	}
+	// Normally the first part after file:// is a hostname, but since
+	// this is often misused we interpret the URL like a normal path
+	// by removing the "file://" from the beginning (if it exists)
+	rawURL = trimPrefixIC(rawURL, "file://")
 
 	if winPathRegex.MatchString(rawURL) {
 		// Windows path
@@ -46,11 +44,7 @@ func Parse(rawURL string) (*url.URL, error) {
 		return url.Parse("file://" + host + strings.ReplaceAll(path, "\\", "/"))
 	}
 
-	u, err := url.Parse(rawURL)
-	if err != nil {
-		return nil, err
-	}
-	return u, nil
+	return url.Parse(rawURL)
 }
 
 // ParseOrPanic parses a url or panics.
@@ -96,4 +90,14 @@ func extractUNCPathParts(uncPath string) (host, path string) {
 		path = "\\" + strings.Join(parts[1:], "\\")
 	}
 	return host, path
+}
+
+// trimPrefixIC returns s without the provided leading prefix string using
+// case insensitive matching.
+// If s doesn't start with prefix, s is returned unchanged.
+func trimPrefixIC(s, prefix string) string {
+	if strings.HasPrefix(strings.ToLower(s), prefix) {
+		return s[len(prefix):]
+	}
+	return s
 }
--- a/urlx/parse.go
+++ b/urlx/parse.go
@@ -13,7 +13,7 @@ var winPathRegex = regexp.MustCompile("^[A-Za-z]:.*")
 
 // Parse parses rawURL into a URL structure with special handling for file:// URLs
 // File URLs with relative paths (file://../file, ../file) will be returned as a
-// utl.URL object without the Scheme set to "file". This is because the file
+// url.URL object without the Scheme set to "file". This is because the file
 // scheme doesn't support relative paths. Make sure to check for
 // both "file" or "" (an empty string) in URL.Scheme if you are looking for
 // a file path.
--- a/urlx/parse.go
+++ b/urlx/parse.go
@@ -2,13 +2,67 @@ package urlx
 
 import (
 	"net/url"
+	"path/filepath"
+	"regexp"
+	"strings"
 
 	"github.com/ory/x/logrusx"
 )
 
+// winPathRegex is a regex for [DRIVE-LETTER]:
+var winPathRegex = regexp.MustCompile("^[A-Za-z]:.*")
+
+// Parse parses rawurl into a URL structure with special handling for file:// URLs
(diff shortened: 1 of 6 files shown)


zepatrik:
I think you could actually have two files and use build tags to distinguish windows and unix. Can you also add the windows github workflow? If the tests for some other packages fail there, we can disable them for this PR.

PatrLind:
Yes, I will look into this. I will apply the fixes to the other PR first and then work on merging them in here together with your suggestion. Wow, this is so involved! I thought it would be a simple fix!

zepatrik:
Haha it's never simple when windows is involved :joy:

urlx/parse.go
```diff
@@ -2,13 +2,60 @@ package urlx
...
 
 	"github.com/ory/x/logrusx"
 )
 
+// winPathRegex is a regex for [DRIVE-LETTER]:
+var winPathRegex = regexp.MustCompile("^[A-Za-z]:.*")
+
+// Parse parses rawURL into a URL structure with special handling for file:// URLs
+// File URLs with relative paths (file://../file, ../file) will be returned as a
+// utl.URL object without the Scheme set to "file". This is because the file
```
zepatrik, line 16:
```suggestion
// url.URL object without the Scheme set to "file". This is because the file
```

PatrLind:
I added build tags now and also added the latest changes from oathkeeper/x

zepatrik (commented):
Can you also add the windows GH workflow? Otherwise I would say this looks perfect :+1:

urlx/parse.go
```diff
@@ -17,7 +58,7 @@ func ParseOrPanic(in string) *url.URL {
 
 // ParseOrFatal parses a url or fatals.
 func ParseOrFatal(l *logrusx.Logger, in string) *url.URL {
-	out, err := url.Parse(in)
+	out, err := Parse(in)
```
aeneasr, line 61: Let's revert this to not break anything in downstream code.

urlx/parse.go
```diff
@@ -2,13 +2,54 @@ package urlx
...
+		return url.Parse("file://" + host + strings.ReplaceAll(path, "\\", "/"))
+	}
+
+	return url.Parse(rawURL)
+}
+
 // ParseOrPanic parses a url or panics.
 func ParseOrPanic(in string) *url.URL {
-	out, err := url.Parse(in)
+	out, err := Parse(in)
```
aeneasr, line 52: Let's revert this to not break anything in downstream code.

urlx/parse.go
```diff
@@ -2,13 +2,54 @@ package urlx
...
+
+// Parse parses rawURL into a URL structure with special handling for file:// URLs
+// File URLs with relative paths (file://../file, ../file) will be returned as a
+// url.URL object without the Scheme set to "file". This is because the file
+// scheme doesn't support relative paths. Make sure to check for
+// both "file" or "" (an empty string) in URL.Scheme if you are looking for
+// a file path.
+// Use the companion function GetURLFilePath() to get a file path suitable
+// for the current operaring system.
+func Parse(rawURL string) (*url.URL, error) {
```
aeneasr, line 22: I would rename this to `ParseWindowsCompatible`

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

urlx/parse.go
PatrLind: This is actually not about parsing URLs for Windows comparability, it is about parsing non standard file URLs. The reason is that Ory uses non standard file URLs in many places throughout the tests and documentation. On windows this caused additional problems since, to get around the non standard URL usage, the code that retrieved the file path didn't use the url.URL.Path field but instead used the url.URL.String() and then just stripped the first `file://` from the beginning. 
Anyway, long story short, This Parse function is written to handle the special case of these non standard file URLs so that they can be parsed in a consistent manner.
However, to make it more simplistic, I think we can drop the UNC support by removing this code:
```
	if strings.HasPrefix(lcRawURL, "\\\\") {
		// Windows UNC path
		// We extract the hostname and create an appropriate file:// URL
		// based on the hostname and the path
		host, path := extractUNCPathParts(rawURL)
		// It is safe to replace the \ with / here because this is POSIX style path
		return url.Parse("file://" + host + strings.ReplaceAll(path, "\\", "/"))
	}
```
Therefore I think it is a better idea to keep the name of this function as Parse() since it has nothing to do with Windows specific parsing.

urlx/parse.go
PatrLind: I understand your concern, what do you think about adding an additional function so that we can have both functions?
`ParseOrFatal2()`?

urlx/parse.go
aeneasr: Well thought out argument - thank you!

aeneasr:
Released as v0.0.156

aeneasr:
Ok, so unfortunately this behaves not as expected in certain cases. `urlx.ParseOrPanic("file://default.schema.json")` incorrectly returns an empty scheme which breaks a couple of upstream tests. I will be moving the new functions to their own function names and we can either fix those cases here or keep them as separate functions.

aeneasr:
What I don't fully understand - maybe that's the problem - why is the scheme empty in this test?

```
{"file://file7.txt", "file7.txt", "file7.txt"},
```

PatrLind:
The reason the scheme is empty is because the file scheme doesn't support relative (non-absolute) paths. The standard library supports relative paths only if the file scheme is empty. So basically what I did was to help the standard library a bit by making the Parse function be able to deal with file:// URLs with only two slashes.

aeneasr:
Ah I see - that makes sense. With open source we learn something new every day :)

The problem is that we rely on `file` scheme being present to determine where we need to load the source from. So I reverted the "breaking" changes in `urlx.*` but left `Parse` intact. We have to see how that works in the rule fetchers in oathkeeper and others as I imagine this breaking in places where relative URLs are used.

PatrLind:
Yes, you are correct, in order to make all tests pass in Oathkeeper I had to add a check for an empty string as well. I will see if Oathkeeper can pass all tests without the ParseOrXXX functions, but last time I tried I had problems. Hopefully there is a way around it.

PR #232, opened 2020-11-13
feat(viperx): Add support for parsing config passed through env vars as json

bmonkman:
## Related issue

@zepatrik as discussed in Slack

## Proposed changes

When setting up Kratos with Helm I would like to be able to configure some of the options via env vars rather than the config file because the values will contain sensitive data like client id / client secret.
This change will try to parse the values that come in to `UnmarshalKey` as JSON, and then falling back to the previous behaviour of encoding then decoding. Further comments below.

## Checklist

      the changes.
- [x] I have added necessary documentation within the code base (if appropriate).

## Further comments

This change will allow me to do something like setting an env var called `SELFSERVICE_METHODS_OIDC` containing a json object that specifies the value of the `oidc` struct. For the most part it shouldn't change the existing functionality around config parsing, except that anything passed via env var that is parseable as JSON now will be actually parsed.
As far as I can think of this shouldn't be a problem for the most part except in cases where someone might be trying to specify quotes or brackets in the value they are passing via env var, for example if they were doing something like `DSN="\"blahblah\"" kratos serve` and expecting to get the literal `"blahblah"`. After this change it would be parsed and the value would be `blahblah`.
Not sure if this edge case is significant.

### DIFF
--- a/viperx/vars.go
+++ b/viperx/vars.go
@@ -28,13 +28,13 @@ func UnmarshalKey(key string, destination interface{}) error {
 			return errors.WithStack(err)
 		}
 
-		// Try decoding the json directly
+		// Try decoding the json directly. If it decodes successfully, return immediately.
 		if err := json.NewDecoder(&b).Decode(destination); err == nil {
 			return nil
 		}
 	}
 
-	// If it's not a string or not valid json, use the value as it was originally provided
+	// If it's not a string or not valid json, use the value as it was originally provided.
 	if err := json.NewEncoder(&b).Encode(mapx.ToJSONMap(value)); err != nil {
 		return errors.WithStack(err)
 	}
--- a/viperx/vars.go
+++ b/viperx/vars.go
@@ -20,8 +20,24 @@ func UnmarshalKey(key string, destination interface{}) error {
 	}
 
 	var b bytes.Buffer
-	if err := json.NewEncoder(&b).Encode(mapx.ToJSONMap(viper.Get(key))); err != nil {
+
+	// This may be a string in the case where a value was provided via an env var.
+	// If it's a string, try to decode it as json.
+	if v, ok := value.(string); ok {
+		if _, err := b.WriteString(v); err != nil {
+			return errors.WithStack(err)
+		}
+
+		// Try decoding the json directly
+		if err := json.NewDecoder(&b).Decode(destination); err == nil {
+			return nil
+		}
+	}
+
+	// If it's not a string or not valid json, use the value as it was originally provided
+	if err := json.NewEncoder(&b).Encode(mapx.ToJSONMap(value)); err != nil {
 		return errors.WithStack(err)
 	}
+
 	return errors.WithStack(json.NewDecoder(&b).Decode(destination))
 }
--- a/viperx/vars_test.go
+++ b/viperx/vars_test.go
@@ -23,3 +23,41 @@ func TestUnmarshalKey(t *testing.T) {
 	assert.EqualValues(t, true, config.Enabled)
 	assert.EqualValues(t, `{"providers":[{"client_id":"kratos-client","client_secret":"kratos-secret","id":"hydra","issuer_url":"http://127.0.0.1:4444/","mapper_url":"file://test/e2e/profiles/oidc/hydra.jsonnet","provider":"generic","scope":["offline"]},{"client_id":"google-client","client_secret":"kratos-secret","id":"google","issuer_url":"http://127.0.0.1:4444/","mapper_url":"file://test/e2e/profiles/oidc/hydra.jsonnet","provider":"generic","scope":["offline"]},{"client_id":"github-client","client_secret":"kratos-secret","id":"github","issuer_url":"http://127.0.0.1:4444/","mapper_url":"file://test/e2e/profiles/oidc/hydra.jsonnet","provider":"generic","scope":["offline"]}]}`, string(config.Config))
 }
+
+func TestUnmarshalKeyWithJsonObject(t *testing.T) {
+	viper.SetConfigFile("./stub/.unmarshal-key.yml")
+	viper.Set("selfservice.strategies.oidc", `{"enabled":true, "config":{"providers": [{"client_id":"a-different-client","client_secret":"a-different-secret","issuer_url":"http://127.0.0.1:4444/","mapper_url":"file://test/e2e/profiles/oidc/a-different-file.jsonnet","provider":"github","scope":["user:email"]}]}}`)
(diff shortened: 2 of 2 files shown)


viperx/vars_test.go
```diff
@@ -23,3 +23,41 @@ func TestUnmarshalKey(t *testing.T) {
 	assert.EqualValues(t, true, config.Enabled)
 	assert.EqualValues(t, `{"providers":[{"client_id":"kratos-client","client_secret":"kratos-secret","id":"hydra","issuer_url":"http://127.0.0.1:4444/","mapper_url":"file://test/e2e/profiles/oidc/hydra.jsonnet","provider":"generic","scope":["offline"]},{"client_id":"google-client","client_secret":"kratos-secret","id":"google","issuer_url":"http://127.0.0.1:4444/","mapper_url":"file://test/e2e/profiles/oidc/hydra.jsonnet","provider":"generic","scope":["offline"]},{"client_id":"github-client","client_secret":"kratos-secret","id":"github","issuer_url":"http://127.0.0.1:4444/","mapper_url":"file://test/e2e/profiles/oidc/hydra.jsonnet","provider":"generic","scope":["offline"]}]}`, string(config.Config))
 }
+
+func TestUnmarshalKeyWithJsonObject(t *testing.T) {
```
zepatrik, line 27: We usually add these test cases using

```go
func TestUnmarshalKey(t *testing.T) {
    t.Run("case=with JSON object", func(t) {
        ...
    })
}
```

Would be awesome if you could do that as well :wink:

viperx/vars.go
```diff
@@ -20,8 +20,24 @@ func UnmarshalKey(key string, destination interface{}) error {
...
+		}
+
+		// Try decoding the json directly
+		if err := json.NewDecoder(&b).Decode(destination); err == nil {
+			return nil
+		}
+	}
+
+	// If it's not a string or not valid json, use the value as it was originally provided
+	if err := json.NewEncoder(&b).Encode(mapx.ToJSONMap(value)); err != nil {
```
zepatrik, line 38: So this looks good but is not that easy to read. Maybe you can change it to be more like:
```go
	switch v := value.(type) {
	case string:
		if _, err := b.WriteString(v); err != nil {
			return errors.WithStack(err)
		}
	default:
		if err := json.NewEncoder(&b).Encode(mapx.ToJSONMap(v)); err != nil {
			return errors.WithStack(err)
		}
	}
	// at this point the buffer holds the encoded value
	return errors.WithStack(json.NewDecoder(&b).Decode(destination))
```

viperx/vars.go
```diff
@@ -20,8 +20,24 @@ func UnmarshalKey(key string, destination interface{}) error {
...
+
+	// This may be a string in the case where a value was provided via an env var.
+	// If it's a string, try to decode it as json.
+	if v, ok := value.(string); ok {
+		if _, err := b.WriteString(v); err != nil {
+			return errors.WithStack(err)
+		}
+
+		// Try decoding the json directly
+		if err := json.NewDecoder(&b).Decode(destination); err == nil {
```
aeneasr, line 32: You could use `github.com/tidwall/gjson.Valid` for a faster and less memory expensive operation here!

zepatrik (changes requested):

aeneasr (commented):
Nice change! Got some ideas :)

viperx/vars.go
zepatrik: Valid would not decode it into `destination` though :thinking:

aeneasr: Ah yeah, I misread the error handling - though it is `err != nil`

viperx/vars.go
bmonkman: Sure thing. I had it as a `switch` before because I was trying a slightly different version but changed it because a switch with only one case and a default felt weird, but I don't mind switching it back if you think it's more readable.

viperx/vars.go
bmonkman: I'll make the comment a bit clearer.

viperx/vars.go
bmonkman: Actually this won't quite work as written because it doesn't account for the fallback. If it's a string but it can't be parsed as JSON we still need to run the `json.NewEncoder(&b).Encode(mapx.ToJSONMap(v))` piece.

zepatrik: OK, my idea was just whether we could move the JSON decode to the end of the function and only have it once for every branch of the program.

bmonkman: I don't think so, because in the case of a string that is not JSON,  `Decode` actually does need to be called twice. Once to realize it's not JSON, and then again after the `Encode` happens.
There are really 3 cases that are being handled here: 
1. The value is of whatever non-string type came in from viper (probably a map) - encode then decode
2. The value is of type `string` and contains JSON - just decode the JSON
3. The value is of type `string` and does not contain JSON - encode then decode (this case seems a bit wasteful as we're just encoding and decoding a string, but it's also necessary because the `Encode` lets it fail properly if it's trying to unmarshal into the wrong type.)

aeneasr (approved):

aeneasr:
Awesome 🎉 

Thank you for your contribution!

zepatrik (approved):

bmonkman:
No problem! 
I'd love to have this change available in the next version of Kratos. Is there anything I can do to help there or do you all handle bumping the version?

zepatrik:
I released it as v0.0.164
It should be enough to bump the dependency right? Would be awesome if you could do that and add a short note to https://github.com/ory/kratos/blob/master/.schema/config.schema.json#L664 saying that the whole key should be set through env vars as one JSON string. It will then automatically be rendered on this docs page: https://www.ory.sh/kratos/docs/reference/configuration/

bmonkman:
The way the docs are currently rendered, they automatically get that "Set this value using environment variables" section if it's not an object which was already problematic because it would add that section for arrays even though they could not be set with an env var. After this change, you can set any of those objects via env var, so should I change that logic slightly to render the comment blocks or just leave it as-is and just add a comment to the OIDC object specifically because there is a more clear use case to specifying it via env var?

aeneasr:
> The way the docs are currently rendered, they automatically get that "Set this value using environment variables" section if it's not an object. After this change, you can set any of those objects via env var, so should I change that logic slightly to render the comment blocks for all of them? That seems a bit verbose. I could also add a field to the schema to allow us to show that block conditionally for objects, then turn it on for the OIDC object specifically because there is a more clear use-case for specifying that one via env var?

Good question, I think you can add it to the description for now!

PR #243, opened 2020-12-08
feat(watcherx): allow requesting current data

zepatrik:
## Related issue

In many cases the data have to be retrieved when starting the watcher as well.

## Proposed changes

This adds a possibility to signal to the watcher that you need it to send the current value as if it changed into the current state.

### DIFF
--- a/configx/provider.go
+++ b/configx/provider.go
@@ -235,14 +235,13 @@ func (p *Provider) addConfigFile(ctx context.Context, path string, k *koanf.Koan
 				cancel()
 				cancel = cancelInner
 				p.runOnChanges(e, nil)
-				close(c)
 				span.Finish()
 				return
 			}
 		}
 	}(c)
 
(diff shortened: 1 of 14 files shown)


configx/provider.go
```diff
@@ -235,14 +235,13 @@ func (p *Provider) addConfigFile(ctx context.Context, path string, k *koanf.Koan
 				cancel()
 				cancel = cancelInner
 				p.runOnChanges(e, nil)
-				close(c)
```
zepatrik, line 238: This caused a panic in some new test case because the receiver closed the channel instead of the sender.

aeneasr (approved):

aeneasr:
Test is running for 15 minutes already on windows - maybe a timeout?

https://github.com/ory/x/pull/243/checks?check_run_id=1516891652

PR #253, opened 2020-12-17
test: fix flaky configx tests

zepatrik:
## Related issue
CI failing all the time

## Proposed Change

A propagated change is now signaled by a channel instead of a wait group. This results in the test blocking until the change propagates, but subsequent events are ignored. They appeared until now because the cleanup of files and closing of the fsnotify watcher were racy.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   nancy: ory/nancy@0.0.9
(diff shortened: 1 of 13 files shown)


zepatrik:
Please note that I ignored all of the newly reported gosec issues. We have to either fix them in a separate PR or think about a solution and fix it here.

aeneasr (approved):

PR #263, opened 2021-01-20
chore: use interfaces instead of testing types

zepatrik:
## Proposed changes

Using `testing.TB` allows to pass either `*testing.T` or `*testing.B`. Using `require.TestingT` will even allow wrapping the original testing struct to allow e.g. concurrent `FailNow()`. We should use the least restricting type possible from now on.


zepatrik:
I assume more changes will come, therefore the draft PR.

PR #297, opened 2021-03-08
fix: handle migration precedence appropriately

aeneasr:
This patch makes sure that specific migrations (e.g. 1234_foobar.sqlite.up.sql) override generic migrations (e.g. 1234_foobar.up.sql).

/cc @zepatrik


zepatrik:
Haha, fixed that in pop some time ago already :wink: https://github.com/gobuffalo/pop/pull/533

PR #306, opened 2021-03-17
feat: validity check for sqlite dsn

Benehiko:
## Proposed changes

Add check for sqlite dsn string allowing the following three formats for in-memory connections
1 - just in memory
`sqlite://file::memory:?_fk=true`
2 - shared connection
`sqlite://file::memory:?_fk=true&cache=shared`
3 - shared but unique in the same process
`sqlite://file:uniquedb:?_fk=true&mode=memory&cache=shared`

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
      and signed the CLA.
- [X] I have read the [security policy](../security/policy).
- [X] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added necessary documentation within the code base (if
      appropriate).

PR #309, opened 2021-03-19
fix: sqlite dsn check

Benehiko:
## Related issue

Related PR https://github.com/ory/x/pull/306

## Proposed changes

Added two more cases to check for:
- `memory`
- `sqlite://:memory:?_fk=true`

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
      and signed the CLA.
- [X] I confirm that this pull request does not address a security
      the changes.
- [X] I have added tests that prove my fix is effective or that my feature
      works.
      appropriate).


dbal/dsn_test.go
```diff
@@ -1,25 +1,38 @@
 package dbal
 
 import (
-	"testing"
-
 	"github.com/stretchr/testify/require"
+	"testing"
```
zepatrik, line 5: `make format`

dbal/dsn_test.go
```diff
@@ -1,25 +1,38 @@
...
+	}
+
+	for k, v := range testCases {
+		t.Logf("testing: %s\nexpected:%v\n", k, v)
+		ok := IsMemorySQLite(k)
+		if v {
+			require.True(t, ok)
+		} else {
+			require.False(t, ok)
+		}
```
zepatrik, line 36: I would say this is more of an `assert` because the test can still continue with the other cases if one fails.
`require` should be used for stuff that is required to continue, e.g. there could be panics otherwise or it is in some other broken state.
```suggestion
		assert.Equal(t, v, ok)
```

dbal/dsn.go
```diff
@@ -15,7 +15,11 @@ const (
 // - shared but unique in the same process
 // see: https://sqlite.org/inmemorydb.html
 func IsMemorySQLite(dsn string) bool {
-	r := regexp.MustCompile(`sqlite://file::?\w+:\?_fk=true&*(cache=shared)?(mode=memory)?`)
+	if dsn == "memory" {
+		return true
+	}
+
+	r := regexp.MustCompile(`sqlite://(file)?::?\w+:\?_fk=true&*(cache=shared)?(mode=memory)?`)
```
zepatrik, line 22: This pattern still does not fully convince me.
If `file` is optional, shoudn't the colon also be optional?
```suggestion
	r := regexp.MustCompile(`sqlite://(file:)?:?\w+:\?_fk=true&*(cache=shared)?(mode=memory)?`)
```
The query parameters are only matching in the order they are specified here, so
```
sqlite://file:uniquedb:?_fk=true&mode=memory&cache=shared
^-- match start                            ^-- match end
sqlite://file:uniquedb:?_fk=true&cache=shared&mode=memory
^-- match start                                         ^-- match end
```
Also, we are missing a `&` in between the query params, and why is it `*` after the first `&`?
```suggestion
	r := regexp.MustCompile(`sqlite://(file)?::?\w+:\?_fk=true(&cache=shared)?(&mode=memory)?`)
```
Also, don't we have to have either the file `memory` or some other file name AND the `mode=memory`?

zepatrik (changes requested):

dbal/dsn_test.go
Benehiko: Using require.Equal(t, v, IsMemorySQLite(k)) rather so that it fails and exits at the point of failure (for test debugging reasons) since this is in a loop.

dbal/dsn.go
Benehiko: refactoring this

Benehiko: so the golang regex engine does not have positive lookahead (which seems to be required for matches in any position).
Do you know of any other methodology? https://stackoverflow.com/questions/4389644/regex-to-match-string-containing-two-names-in-any-order/4389683#4389683

dbal/dsn_test.go
zepatrik: For debugging you can use the `-failfast` flag when executing tests.

dbal/dsn.go
zepatrik: Either `...(&cache=shared)?(&mode=memory)?(&cache=shared)?` or `url.Parse` and ditch the regex completely.

zepatrik: Or split the string into query and rest, do a regex on rest and parse the query string to do checks on that.

dbal/dsn_test.go
```diff
@@ -3,23 +3,33 @@ package dbal
...
-	require.False(t, IsMemorySQLite("sqlite://file::?_fk=true"))
-	require.False(t, IsMemorySQLite("sqlite://file:::?_fk=true"))
-	require.False(t, IsMemorySQLite("postgresql://username:secret@localhost:5432/database"))
+	testCases := map[string]bool{
+		SQLiteInMemory:               true,
+		SQLiteSharedInMemory:         true,
+		"memory":                     true,
+		"sqlite://:memory:?_fk=true": true,
+		"sqlite://file:uniquedb:?_fk=true&mode=memory":              true,
+		"sqlite://file:uniquedb:?_fk=true&cache=shared":             true,
```
zepatrik, line 16: This is not a valid memory database according to https://sqlite.org/inmemorydb.html
```suggestion
		"sqlite://file:uniquedb:?_fk=true&cache=shared":             false,
```

dbal/dsn_test.go
```diff
@@ -3,23 +3,33 @@ package dbal
...
+		"sqlite://file:uniquedb:?_fk=true&mode=memory&cache=shared": true,
+		"sqlite://file:uniquedb:?_fk=true&cache=shared&mode=memory": true,
+		"sqlite://file:uniquedb:?mode=memory":                       true,
+		"sqlite://file:::uniquedb:?_fk=true&mode=memory":            false,
```
zepatrik, line 20: This is a valid memory database according to https://sqlite.org/inmemorydb.html
```suggestion
		"sqlite://file:::uniquedb:?_fk=true&mode=memory":            true,
```

dbal/dsn.go
```diff
@@ -15,11 +15,6 @@ const (
...
 func IsMemorySQLite(dsn string) bool {
-	r := regexp.MustCompile(`sqlite://file::?\w+:\?_fk=true&*(cache=shared)?(mode=memory)?`)
-
-	if r.MatchString(dsn) {
-		return true
-	}
-
-	return false
+	r := regexp.MustCompile(`^(?:sqlite://((file:((:memory)|(\w+)):)|(:memory:))(\?.*))$|(?:^(memory|:memory:)$)`)
+	return r.MatchString(dsn)
```
zepatrik, line 19: OK sorry to bother again, but I still see problems with this.
Why not solely rely on the rules presented in https://sqlite.org/inmemorydb.html ?
```suggestion
	if dsn == "memory" {
		return true
	}

	u, err := url.Parse(dsn)
	if err != nil {
		return false
	}

	return u.Scheme == "sqlite" && (u.Host == "file::memory:" || u.Host == ":memory:" || (u.Query().Get("mode") == "memory" && strings.HasPrefix(u.Host, "file:")))
```

The current pattern does not accept e.g. `sqlite://file:memdb1?mode=memory&cache=shared` (literally taken from the sqlite docs).

dbal/dsn_test.go
```diff
@@ -3,23 +3,33 @@ package dbal
...
+		"sqlite://?_fk=true&cache=shared":                      false,
+		"sqlite://file::?_fk=true":                             false,
+		"sqlite://file:::?_fk=true":                            false,
+		"postgresql://username:secret@localhost:5432/database": false,
+	}
+
+	for k, v := range testCases {
+		t.Logf("testing: %s\nexpected:%v\n", k, v)
+		assert.Equal(t, v, IsMemorySQLite(k))
+	}
```
zepatrik, line 34: Just ran the tests locally, it is way nicer and easier to debug if you have subtests for each case in the loop:
```suggestion
	for dsn, expected := range testCases {
		t.Run("dsn="+dsn, func(t *testing.T) {
			assert.Equal(t, expected, IsMemorySQLite(dsn))
		})
	}
```

zepatrik (changes requested):
Ok I think the regex approach is not a good solution in general. Why not parse the URL and check it's properties?

Benehiko:
Please be merciful :pray:

zepatrik (approved):
Awesome, looks good 👍

dbal/dsn.go
```diff
@@ -15,11 +15,6 @@ const (
...
 // see: https://sqlite.org/inmemorydb.html
 func IsMemorySQLite(dsn string) bool {
-	r := regexp.MustCompile(`sqlite://file::?\w+:\?_fk=true&*(cache=shared)?(mode=memory)?`)
-
-	if r.MatchString(dsn) {
-		return true
-	}
-
-	return false
+	r := regexp.MustCompile(`^(sqlite://file:(?:.+)\?((\w+=\w+)(&\w+=\w+)*)?(&?mode=memory)(&\w+=\w+)*)$|(?:sqlite://(file:)?:memory:(?:\?\w+=\w+)?(?:&\w+=\w+)*)|^(?:(?::memory:)|(?:memory))$`)
```
aeneasr, line 18: This can be a global:

```
var dsnRegex = regexp.MustCompile(...)
```

aeneasr (changes requested):

PR #310, opened 2021-03-22
fix(clidoc): sidebar spam

zepatrik:
Regression of the sidebar spam bug. Looks like this was caused because of a change to the sidebar file format. Added a snapshot test.

### DIFF
--- a/clidoc/generate.go
+++ b/clidoc/generate.go
@@ -16,6 +16,8 @@ import (
 	"github.com/tidwall/sjson"
 )
 
+const sideBarLabel = "Command Line Interface (CLI)"
+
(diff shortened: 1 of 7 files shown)


aeneasr:
Tests fail

zepatrik:
Right, this is because of https://github.com/tidwall/pretty/issues/11

zepatrik:
Ah no it was because `filepath` was used instead of `path` (which failed the windows test). Linux was just flaky.

zepatrik:
And also `\r\n` line endings :sob:

aeneasr (approved):

Issue #313, opened 2021-03-29
logrusx/helper.go trace logging not working


aarmam:
**Describe the bug**
Switch from go.opentelemetry.io/otel v0.6.0 to go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace v0.13.0 breaks logging in https://github.com/ory/x/blob/master/logrusx/helper.go#L84-L93

**To Reproduce**
Request with header 
traceparent=00-b67866858ff865fb4d89725465fd2bf4-e7f3f812960b1654-01
should log "otel":{"span_id":"e7f3f812960b1654","trace_id":"b67866858ff865fb4d89725465fd2bf4"}

I tried to switch back to go.opentelemetry.io/otel v0.6.0 and then it worked.

aeneasr:
So what could fix that?

wangleibeckham:
This is still an issue when upgrading opentelemetry to 0.20.0 as [spanContext](https://github.com/open-telemetry/opentelemetry-go/blob/main/trace/trace.go#L205) now has only protected properties. Which makes [here](https://github.com/ory/x/blob/master/logrusx/helper.go#L90) and [here](https://github.com/ory/x/blob/master/logrusx/helper.go#L93) to break.

I can create a PR for this. @aeneasr let me know your thoughts. Thanks!

PR #315, opened 2021-03-31
Use TraceContext propagator that supports the W3C Trace Context format

aarmam:
## Related issue

https://github.com/ory/x/issues/313

Caution I'm not go developer!


aeneasr:
Thank you, this looks great! Could you please sign the CLA (just two clicks), then we can merge it right away!

PR #368, opened 2021-07-02
fix(configx): more locks to prevent race conditions

zepatrik:
## Related issue

Got some race failures in Keto when running `go test -race`.

## Proposed changes

Add locks to places where koanf is accessed and written to.

### DIFF
--- a/configx/provider.go
+++ b/configx/provider.go
@@ -46,7 +46,7 @@ type tuple struct {
 }
 
 type Provider struct {
-	l sync.Mutex
+	l sync.RWMutex
 	*koanf.Koanf
 	immutables []string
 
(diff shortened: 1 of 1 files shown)


configx/provider.go
```diff
@@ -333,6 +337,9 @@ func (p *Provider) Set(key string, value interface{}) error {
 }
 
 func (p *Provider) BoolF(key string, fallback bool) bool {
+	p.l.RLock()
```
aeneasr, line 340: Are these really needed? I was under the impression that koanf reading is lock free. Or is the problem the setters?

zepatrik: Exactly, when we replace the koanf instance we get a race condition. https://app.circleci.com/pipelines/github/ory/keto/1308/workflows/238f2513-8e2e-4182-b911-5fdf94c19658/jobs/8684?invite=true#step-106-311

PR #369, opened 2021-07-06
feat: merged file system

zepatrik:
This adds a small utility `fs.FS` that implements a merged file system. Further, the types in `popx.Migrator` are updated so that any `fs.FS` can be used. This allows us to do ` popx.NewMigrationBox(fsx.Merge(migrations, networkx.Migrations), popx.NewMigrator(p.Connection(ctx), p.d.Logger(), nil, 0))`, which in turn properly returns the status and also implements down migrations. Marked the `networkx.MigrateUp` func as deprecated, but maybe rather delete right away?

### DIFF
--- a/fsx/merge.go
+++ b/fsx/merge.go
@@ -1,0 +1,30 @@
+package fsx
+
+import (
+	"io/fs"
+
+	"github.com/pkg/errors"
+)
+
+type merged []fs.FS
+
+var (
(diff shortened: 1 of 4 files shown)


aeneasr:
Cool idea 👍

aeneasr (approved):

zepatrik:
Tests are failing because of flaky watcherx.... gonna merge :grin:

PR #370, opened 2021-07-06
fix(fsx): make merged file system walkable and fully comply with "testing/fstest"

zepatrik:
## Related issue

The initial implementation was not `fs.Walk`able. This now implements all features and applies the full `fstest` testsuit. This should guarantee that the filesystem works.
Please note that we can't return `io.EOF` wrapped with `errors.WithStack` because of https://github.com/golang/go/issues/47062

### DIFF
--- a/fsx/merge.go
+++ b/fsx/merge.go
@@ -1,30 +1,218 @@
 package fsx
 
 import (
+	"io"
 	"io/fs"
+	"sort"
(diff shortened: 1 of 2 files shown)


aeneasr (commented):
lgtm

aeneasr (approved):

PR #373, opened 2021-07-13
fix(castx): use csv parsing for string slices

zepatrik:
related to https://github.com/ory/hydra/issues/2609

### DIFF
--- a/castx/castx.go
+++ b/castx/castx.go
@@ -1,12 +1,10 @@
 package castx
 
 import (
+	"encoding/csv"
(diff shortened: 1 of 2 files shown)


zepatrik:
Maybe we should make those changes directly in configx instead because castx might be used in other places and this is a breaking change.

aeneasr (approved):

aeneasr:
https://github.com/search?q=org%3Aory+ToStringSlice&type=code

PR #374, opened 2021-07-15
fix: allow to override some nested values in config.

harnash:
Currently it is not possible to override some values in the config if they are nested such as `SELFSERVICE_METHODS_OIDC_CONFIG_PROVIDERS`. It is still not possible to override some values that are nested inside `array` but this at least should allow to replace the whole array with some serialized JSON struct. It is not ideal but it should have some more options to deal with injecting secrets into Kratos et al. 

## Related issue(s)
This is a follow up on https://github.com/ory/kratos/issues/1535

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

My initials investigation showed that even if I can generate nested keys for elements inside JSON arrays it is still hard to (impossible? https://github.com/knadh/koanf/issues/74) to make koanf to properly set those variables into the config:
```
I[#/selfservice/methods/oidc/config/providers] S[#/properties/selfservice/properties/methods/properties/oidc/properties/config/properties/providers/type] expected array, but got string
```
For reference Viper seems to support such cases (see: https://github.com/spf13/viper#accessing-nested-keys) and I know Viper is different kind of beast and has it's own set of problems but perhaps it is wise to look for alternative? Another option might be to use combined config that are merged into one (still not ideal but would help with secrets handling?).

### DIFF
--- a/configx/koanf_env.go
+++ b/configx/koanf_env.go
@@ -42,8 +42,16 @@ func NewKoanfEnv(prefix string, schema []byte) (*env.Env, error) {
 	}
 
 	decode := func(value string) (v interface{}) {
-		_ = json.Unmarshal([]byte(value), v)
-		return v
+		b := []byte(value)
+		var arr []interface{}
+		if err := json.Unmarshal(b, &arr); err == nil {
+			return &arr
+		}
+		h := map[string]interface{}{}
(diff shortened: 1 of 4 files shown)


aeneasr (approved):
Awesome, thank you!

PR #377, opened 2021-07-20
feat(tracing): add tracing config schema

zepatrik:
From now on we can use `"$ref": "ory://tracing-config"` in the config schemas instead of having to copy it around.
See related https://github.com/ory/keto/issues/662


aeneasr:
That is unfortunately not a good idea. I have done this mistake in Ory Oathkeeper before. Here are the reasons:

1. You are referencing a non-versioned URL. This will eventually update, be refactored, changed. All older builds will no longer work as soon as this changes. If you want to reference a URL, it must be versioned (like a git hash).
2. You are referencing a HTTP URL. This means that Ory Keto will not work in envs without network access. We got hit by this in Ory Oathkeeper as well

If you want to centralize the config, it needs to be ensured that it uses `embed` somehow, and in particular not referencing any URL, in order to get reproducible builds.

Otherwise, this LGTM afaict

tracing/config.go
```diff
@@ -1,23 +1,55 @@
...
+// The interface is specified instead of `jsonschema.Compiler` to allow the use of any jsonschema library fork or version.
+func AddConfigSchema(c interface {
+	AddResource(url string, r io.Reader) error
+}) error {
+	schema, err := ConfigSchema.ReadFile("config.schema.json")
+	if err != nil {
+		panic("unrecoverable error: could not read file from embed.FS: " + err.Error())
+	}
+
+	return c.AddResource(ConfigSchemaID, bytes.NewBuffer(schema))
```
zepatrik, line 54: > 1. You are referencing a non-versioned URL. This will eventually update, be refactored, changed. All older builds will no longer work as soon as this changes. If you want to reference a URL, it must be versioned (like a git hash).
> 2. You are referencing a HTTP URL. This means that Ory Keto will not work in envs without network access. We got hit by this in Ory Oathkeeper as well

The URL is just a random placeholder., it only has to be unique The schema is loaded from the `embed.FS`, versioned by the version of ory/x you have in the `go.mod`. That is why we have to add it to the compiler here @aeneasr 

We can instead use e.g. the string `ory/schemas/tracing` or just anything we can assume to be unique.

aeneasr: yeah, a fixed string will probably prevent the loader from trying to load it from remote if it's not available for some reason (e.g. build error) :)

PR #379, opened 2021-07-22
feat(prometheus): Add additional Prometheus metrics to instrument Handlers

t-tomalak:
Current metrics are very limited and show only response time buckets per handler. To properly monitor service we also should have information about status codes, response sizes, number of requests etc. Therefore we should add additional ones. Promhttp has its own request duration instrumentation, and I think that we should use this one instead of the custom handler created for `_response_time_seconds` with a timer. Also, with this rewrite for prometheusx, it will be easier for the others to add additional metrics if needed in the future.

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).


aeneasr (changes requested):
Thank you for the contribution! I think this is going into the right direction.

It would be awesome if you could add tests! :relieved:

Tests are important to open source projects because the allow reviewers to better judge if the changes are doing what was intended. They also ensure that your changes never regress / break when someone else works on code that is related to this.

If you need help with the tests feel free to push your work in progress and ping myself or another maintainer :raising_hand:

t-tomalak:
@aeneasr Added tests. I only check for the metrics' existence and if the "most important" label is included, should I add more tests or do we think that this is enough?

aeneasr:
Thank you! I added some changes but overall this looks very good! Ready to be merged :)

aeneasr (approved):

aeneasr:
Whoops, mixed up the tabs, sorry! :D

harnash:
@aeneasr Is this PR approved and ready for merge or there is still something to address here? Tests seem to fail randomly.

prometheusx/metrics.go
```diff
@@ -1,32 +1,69 @@
...
+	duration        *prometheus.HistogramVec
+	responseSize    *prometheus.HistogramVec
+	requestSize     *prometheus.HistogramVec
+	handlerStatuses *prometheus.CounterVec
 }
 
 // Method for creation new custom Prometheus  metrics
 func NewMetrics(app, version, hash, date string) *Metrics {
+	labels := map[string]string{
+		"version":   version,
```
aeneasr, line 26: Would it maybe make sense to add the `app` to the label instead of prefixing it?

aeneasr (commented):
Sorry for the late reply! Thank you for the tests, I have a general question regarding naming / labeling. Do you maybe have some best practices which I could look up for this as well?

prometheusx/metrics.go
t-tomalak: It would make sense, but I didn't want to break previous metrics names as someone could already use it

aeneasr: Ok cool, then I propose to change it right away :)

Do you maybe have a doc explaining what some best practices are for these type of metrics?

t-tomalak: Sure, I will update it then. For the metrics, we wanna follow and deliver basic "RED" metrics, were we can found the Rate, Error and Duration. This is example article, but I'm sure that you could find more https://grafana.com/blog/2018/08/02/the-red-method-how-to-instrument-your-services/.

prometheusx/metrics.go
```diff
@@ -1,32 +1,70 @@
...
-					"buildTime": date,
-				},
+				Name:        "response_time_seconds",
+				Help:        "Description",
+				ConstLabels: labels,
 			},
 			[]string{"endpoint"},
 		),
+		totalRequests: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name:        "requests_total",
```
aeneasr, line 41: Since some services support different API protocols (REST, gRPC) I think these metrics should be prefixed with `http_` or (for gRPC) `grpc_`

aeneasr (changes requested):

prometheusx/metrics.go
t-tomalak: Hello, sorry for my late response. I'm not an expert about gRPC, but could you give me an example wherein Kratos (or another ory service) we are using this instrumentation? Because I think that for the gRPC handler there should be different metrics manager as for example statuses are different, and probably logic for handlers itself?

aeneasr: Hey there, sorry for my late reply too :) We are using gRPC in Ory Keto and are planning on adding it to more services at some point. It is true that not everything overlaps (e.g. status codes), but some things do overlap like requests_total. I saw in the prometheus docs and somewhere else that prefixing the protocol is a good idea, which is why I was asking about this :)

harnash: Will 8610e55 suffice?

aeneasr: Absolutely! :)

aeneasr (approved):
Awesome, thank you ! :)

PR #381, opened 2021-07-27
chore: remove old code that introduced a dependency to packr

zepatrik:
Required to resolve security warnings like https://app.circleci.com/pipelines/github/ory/keto/1362/workflows/47e794cd-a369-4435-9140-4e1615477244/jobs/8952


.github/workflows/test.yml
```diff
@@ -62,7 +62,7 @@ jobs:
           sudo cp -i cockroach-v21.1.2.linux-amd64/cockroach /usr/local/bin/
           rm -rf cockroach-v21.1.2.linux-amd64
       - name: Prepare nancy dependency list
-        run: go list -json -m all > go.list
+        run: go list -json all > go.list
```
aeneasr, line 65: Revert this, it will actually break vuln checking because:

```
$ go list -json all
go: warning: "all" matched no packages
```

not sure why it got there in the first place.

aeneasr (changes requested):

.github/workflows/test.yml
zepatrik: Problem is that https://github.com/ory/x/runs/3173594979 was failing although 

```
go mod why github.com/gobuffalo/packr
# github.com/gobuffalo/packr
(main module does not need package github.com/gobuffalo/packr)
```

I assume because packr (for a unknow reason to me) is still in the `go.sum`.

Btw, I get 

```
$ go list -json all | head
{
        "Dir": "/usr/lib/go/src/archive/tar",
        "ImportPath": "archive/tar",
        "Name": "tar",
        "Doc": "Package tar implements access to tar archives.",
        "Target": "/usr/lib/go/pkg/linux_amd64/archive/tar.a",
        "Root": "/usr/lib/go",
        "Match": [
                "all"
        ],
```

and I think this also happens on the CI because otherwise there would be an error about malformed JSON.
Can't really find resources on what to use TBH....

zepatrik: Investigated a bit more, and the problem is other ory packages required in this repo, that require older versions of `ory/x`, e.g. 

- github.com/ory/analytics-go/v4@v4.0.0 github.com/ory/x@v0.0.110
- github.com/ory/herodot@v0.9.2 github.com/ory/x@v0.0.127
- github.com/ory/jsonschema/v3@v3.0.1 github.com/ory/x@v0.0.93
- github.com/ory/herodot@v0.7.0 github.com/ory/x@v0.0.84

so we basically have to merge this with the failing CI to then update all the dependencies and prune old versions...

zepatrik: I guess further it would be a good idea to not have these circular dependencies....

aeneasr: Which go version do you have?

zepatrik:
```
go version
go version go1.16.5 linux/amd64
```

aeneasr: Ok, removing `-m` toggle was introduced in the latest versions (I think 1.16.4) and was not available before. What is the reason for omitting it? And have you ensured that the CI is using an image that has the new behavior?

aeneasr: And also the reason for omitting it :)

zepatrik: As explained above we have circular dependencies to older versions of `ory/x`. Without the `-m` flag, it seems, only direct dependencies are listed. At least it will not include `packr`, while the `-m` option will. We either merge this with the failing check, propagate the update to the other Ory repos, and then bump them here again. Or only check without `-m`. That is the part I referred to as

> Can't really find resources on what to use TBH

aeneasr: I see, so deep dependency checking is required for proper vulnerability detection as it would be possible that we use library foo which uses library leftpad with a serious vulnerability. I believe omitting `-m` would not find that.

The approach with bumping other deps is acceptable / desirable!

PR #399, opened 2021-09-28
feat: gRPC telemetry interceptors

zepatrik:
## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/metricsx/middleware.go
+++ b/metricsx/middleware.go
@@ -21,9 +21,13 @@
 package metricsx
 
 import (
+	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"github.com/ory/x/configx"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/status"
 	"net"
 	"net/http"
 	"net/url"
@@ -31,8 +35,6 @@ import (
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/ory/x/configx"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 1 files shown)


metricsx/middleware.go
```diff
@@ -298,9 +341,9 @@ func (sw *Service) anonymizePath(path string, salt string) string {
 
 	for _, p := range paths {
 		p = strings.ToLower(p)
-		if len(path) == len(p) && path[:len(p)] == strings.ToLower(p) {
+		if path == p {
 			return p
-		} else if len(path) > len(p) && path[:len(p)+1] == strings.ToLower(p)+"/" {
+		} else if strings.HasPrefix(path, p) {
```
zepatrik, line 346: I think the previous code was a bit complex here, but this should to exactly the same :sweat_smile:

metricsx/middleware.go
```diff
@@ -288,6 +292,45 @@ func (sw *Service) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.
...
+
+	resp, err := handler(ctx, req)
+
+	if sw.optOut {
+		return resp, err
+	}
+
+	latency := time.Since(start) / time.Millisecond
+
+	if err := sw.c.Enqueue(analytics.Page{
```
aeneasr, line 309: Can `info` contain any sensitive information like UUIDs or similar?

zepatrik: It only has two fields:
```go

// UnaryServerInfo consists of various information about a unary RPC on
// server side. All per-rpc information may be mutated by the interceptor.
type UnaryServerInfo struct {
	// Server is the service implementation the user provides. This is read-only.
	Server interface{}
	// FullMethod is the full RPC method string, i.e., /package.service/method.
	FullMethod string
}
```
which are not sensitive as far as I can tell. The `FullMethod` does not contain parameters.

PR #401, opened 2021-10-01
refactor(popx): test migrator should use fs.FS

zepatrik:
This is a breaking change, but can easily be migrated by replacing
```diff
- popx.NewTestMigrator(t, c, "../migrations/sql", "./testdata", l)
+ popx.NewTestMigrator(t, c, os.DirFS("../migrations/sql"), os.DirFS("./testdata"), l)
```

### DIFF
--- a/popx/migrator_test.go
+++ b/popx/migrator_test.go
@@ -58,6 +58,8 @@ func TestMigratorUpgrading(t *testing.T) {
 
 	for name, c := range connections {
 		t.Run(fmt.Sprintf("database=%s", name), func(t *testing.T) {
+			t.SkipNow()
+
 			legacy, err := pkgerx.NewMigrationBox("/popx/stub/migrations/legacy", c, l)
 			require.NoError(t, err)
 			require.NoError(t, legacy.Up())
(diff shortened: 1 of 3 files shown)


aeneasr (approved):

popx/migrator_test.go
```diff
@@ -58,6 +58,8 @@ func TestMigratorUpgrading(t *testing.T) {
 
 	for name, c := range connections {
 		t.Run(fmt.Sprintf("database=%s", name), func(t *testing.T) {
+			t.SkipNow()
```
zepatrik, line 61: This test does not actually test whether the schemas are the same. Dumping of schemas fails because of various reasons (e.g. the statement changed in CRDB, sqlite tries to use the `sqlite` binary). It probably passed until now because the dump function just returned `nil` right away.
In the very end of this test, where we have `assert.EqualValues(t, expected, actual)`, we don't compare the schema dumps, but the errors returned by dump. Therefore I chose to skip the test and not fix it, as it is fundamentally flawed and did not work in the first place.

PR #441, opened 2021-12-17
feat: add base64,https,file fileloader

aeneasr:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/osx/file.go
+++ b/osx/file.go
@@ -1,0 +1,144 @@
+package osx
+
+import (
+	"encoding/base64"
+	"io"
+	"net/url"
+	"os"
+
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/pkg/errors"
+
+	"github.com/ory/x/httpx"
+)
+
+type options struct {
+	disableFileLoader   bool
+	disableHTTPLoader   bool
+	disableBase64Loader bool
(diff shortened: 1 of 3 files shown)


osx/file.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/pkg/errors"
+
+	"github.com/ory/x/httpx"
+)
+
+type options struct {
+	disableFileLoader   bool
+	disableHTTPLoader   bool
+	disableBase64Loader bool
```
zepatrik, line 18: I would make those opt-in and not opt-out.

zepatrik (changes requested):
Maybe we can also have a file-loader instance that keeps the same options around?

osx/file.go
aeneasr: > I would make those opt-in and not opt-out.

I set it explicitly to enabled because otherwise you'd always have to allow every single one. We usually load all of them, only in few instances we need some of them.

> Maybe we can also have a file-loader instance that keeps the same options around?

I think this is a good idea, if you have e.g. a load that is disabled everything per default.

PR #447, opened 2021-12-28
fix: support setting nested array items from envs

aeneasr:
@splaunov this resolves the issue of loading config arrays from the environment. However, it appears that there is a bug in Koanf which overrides existing arrays instead of merging them (see https://github.com/knadh/koanf/issues/125).

If resolved, I think we can merge this and close #420, and have a good solution for providing sensitive values in config items which are defined as arrays! :)

The code is a bit hacky but I think it get's us pretty far. I might have missed an edge case or two but for standard use it should be enough.

Closes #420


splaunov:
This should work but means that user will need to sync numbering of env variables with positions in the config's array. This is brittle.
As for me, it would be better to interpolate arbitrary named variables as in #420.

aeneasr:
Thank you for the feedback! I think the env var substitution is a cool idea! For our cloud system however it will be difficult because it would allow arbitrary users to add configs with substitution that might leak secrets. For that reason, I'd like to have a different approach to it. But if we fail to get the tests passing, I'll merge the substitution one. Maybe we can enable the feature with a flag :)

PR #467, opened 2022-02-11
feat(configx): allow disabling env loader

zepatrik:


aeneasr (approved):

aeneasr:
Or maybe, add a small test?

aeneasr:
Given that this is security relevant...

configx/options_test.go
```diff
@@ -0,0 +1,25 @@
...
+	t.Run("case=does not load env if disabled", func(t *testing.T) {
+		schema := `{"type": "object", "properties": {"path": {"type": "string"}}}`
+
+		envP, err := New(ctx, []byte(schema))
+		require.NoError(t, err)
+		assert.NotZero(t, envP.String("path"))
+
+		nonEnvP, err := New(ctx, []byte(schema), DisableEnvLoading())
+		require.NoError(t, err)
+		assert.Nil(t, nonEnvP.Get("path"))
```
zepatrik, line 23: First we check that it works without the option, and then that it does not if we disable it.

PR #469, opened 2022-02-20
feat: websocket support

Benehiko:
I have come across an issue while working on SPA integration using the [Ory Cli](https://github.com/ory/cli).
When running a development server when working on SPAs the browser opens a websocket connection to the development server for debugging purposes. I noticed that the Ory Cli was dropping the connection with the error:

```
internal error: 101 switching protocols response with non-writable body
```

After further investigation, it seemed that we were returning an `io.NopCloser` and replacing the `response.Body` entirely, which means any modification done internally by the `httputil.ReverseProxy` would be overwritten.

This change allows us to use the websocket response instead with its internal modifications with our modified body.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/proxy/rewrites.go
+++ b/proxy/rewrites.go
@@ -3,18 +3,24 @@ package proxy
 import (
 	"bytes"
 	"compress/gzip"
+	"github.com/pkg/errors"
 	"io"
 	"net/http"
 	"net/url"
 	"path"
 	"strings"
-
-	"github.com/pkg/errors"
 )
 
 type compressableBody struct {
 	buf bytes.Buffer
 	w   io.WriteCloser
+}
+
+// we require a read and write for websocket connections
+var _ io.ReadWriteCloser = new(compressableBody)
+
+func (b *compressableBody) Close() error {
+	return nil
 }
 
 func (b *compressableBody) Write(d []byte) (int, error) {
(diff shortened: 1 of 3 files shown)


proxy/rewrites.go
```diff
@@ -3,20 +3,26 @@ package proxy
...
 type compressableBody struct {
 	buf bytes.Buffer
 	w   io.WriteCloser
 }
 
+// we require a read and write for websocket connections
+var _ io.ReadWriteCloser = new(compressableBody)
+
+func (b *compressableBody) Close() error {
+	return nil
```
aeneasr, line 23: Shouldn't this close the inner elements as to avoid memory / handler leaks? i.e.

```
b.w.Close()
b.buf.Clean() // or whatever the method is called the empties a buffer :)
```

Benehiko: I'll look into this :)

Benehiko: The only thing I could think of was to reset the buffer entirely, but it has no `Close()` function.

```go
func (b *compressableBody) Close() error {
	if b != nil {
		b.buf.Reset()
	}
	return nil
}
```

Benehiko: I added a close for this, we initially returned a https://pkg.go.dev/io#NopCloser.

aeneasr (approved):

PR #475, opened 2022-03-07
feat: support 'leak_sensitive_values' in URL params

nipsufn:
This change ensures that URL query string is not written to logs when `Logf` family of functions is used and `leakSensitive` option is disabled.
URL query can contain sensitive data, e.g. API keys.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/logrusx/helper.go
+++ b/logrusx/helper.go
@@ -101,18 +101,18 @@ func (l *Logger) WithRequest(r *http.Request) *Logger {
 
 func (l *Logger) Logf(level logrus.Level, format string, args ...interface{}) {
 	if !l.leakSensitive {
-		var myArgs []interface{}
-		for _, arg := range args {
-			urlArg, ok := arg.(*url.URL)
-			if ok {
+		for i, arg := range args {
+			switch urlArg := arg.(type) {
+			case url.URL:
 				urlCopy := url.URL{Scheme: urlArg.Scheme, Host: urlArg.Host, Path: urlArg.Path}
-				myArgs = append(myArgs, &urlCopy)
-			} else {
-				myArgs = append(myArgs, arg)
+				args[i] = urlCopy
+			case *url.URL:
+				urlCopy := url.URL{Scheme: urlArg.Scheme, Host: urlArg.Host, Path: urlArg.Path}
+				args[i] = &urlCopy
+			default:
+				continue
 			}
 		}
-		l.Entry.Logf(level, format, myArgs...)
-		return
 	}
 	l.Entry.Logf(level, format, args...)
 }
--- a/logrusx/helper.go
+++ b/logrusx/helper.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


logrusx/helper.go
```diff
@@ -98,6 +99,62 @@ func (l *Logger) WithRequest(r *http.Request) *Logger {
...
 }
 
+func (l *Logger) Logf(level logrus.Level, format string, args ...interface{}) {
+	if !l.leakSensitive {
+		var myArgs []interface{}
+		for _, arg := range args {
+			urlArg, ok := arg.(*url.URL)
+			if ok {
+				urlCopy, _ := url.Parse(urlArg.String())
+				urlCopy.RawQuery = `Value is sensitive and has been redacted. To see the value set config key "log.leak_sensitive_values = true" or environment variable "LOG_LEAK_SENSITIVE_VALUES=true".`
```
zepatrik, line 109:
```suggestion
				urlCopy.RawQuery = l.maybeRedact(urlCopy.RawQuery)
```

logrusx/helper.go
```diff
@@ -98,6 +99,62 @@ func (l *Logger) WithRequest(r *http.Request) *Logger {
 	return ll
 }
 
+func (l *Logger) Logf(level logrus.Level, format string, args ...interface{}) {
+	if !l.leakSensitive {
+		var myArgs []interface{}
+		for _, arg := range args {
+			urlArg, ok := arg.(*url.URL)
+			if ok {
+				urlCopy, _ := url.Parse(urlArg.String())
```
zepatrik, line 108:
```suggestion
				urlCopy := *urlArg
```
The only field that will not be properly copied by this is the userinfo, and I suggest dropping that anyways.

zepatrik (changes requested):
In general, the query is not logged by the middleware right? Wouldn't it be better to just not log the URL manually? I can imagine that sometimes you actually want to log the whole URL.
Anyways, if we stick with this I suggest we apply the same rules as with the request middleware, i.e. only allow `scheme`, `host` and `path`.

logrusx/helper.go
```diff
@@ -98,6 +99,61 @@ func (l *Logger) WithRequest(r *http.Request) *Logger {
...
+		var myArgs []interface{}
+		for _, arg := range args {
+			urlArg, ok := arg.(*url.URL)
+			if ok {
+				urlCopy := url.URL{Scheme: urlArg.Scheme, Host: urlArg.Host, Path: urlArg.Path}
+				myArgs = append(myArgs, &urlCopy)
+			} else {
+				myArgs = append(myArgs, arg)
+			}
+		}
```
zepatrik, line 113: One last optimization we can do, just FYI:
```suggestion
		for i, arg := range args {
			urlArg, ok := arg.(*url.URL)
			if ok {
				urlCopy := url.URL{Scheme: urlArg.Scheme, Host: urlArg.Host, Path: urlArg.Path}
				arg[i] = &urlCopy
			}
		}
```

zepatrik: And maybe even support both `url.URL` and `*url.URL` for the sake of completeness:
```suggestion
		for i, arg := range args {
			switch urlArg := arg.(type) {
			case url.URL:
				urlCopy := url.URL{Scheme: urlArg.Scheme, Host: urlArg.Host, Path: urlArg.Path}
				arg[i] = urlCopy
			case *url.URL:
				urlCopy := url.URL{Scheme: urlArg.Scheme, Host: urlArg.Host, Path: urlArg.Path}
				arg[i] = &urlCopy
			default:
				continue
			}
		}
```

zepatrik (changes requested):

PR #479, opened 2022-03-11
feat: add otelx package

icyphox:
otelx wraps around go.opentelemetry.io/otel to simplify creating and using tracers. This is separate from the ory/x/tracing package as that supports a few more tracers (Zipkin, Instana, ...) which otelx does not, yet.

A note about the tests: the test for Jaeger spans simply spins up a UDP server and passes when data is received. While the byte data looks to be correct, I have not added code to unmarshal it. The reasoning behind this decision:
- OpenTelemetry's compact Thrift generation is already tested
- github.com/apache/thrift/lib/go/thrift is fairly complex and would take a lot of time to get setup, just for one test

https://github.com/ory-corp/cloud/issues/2191

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).


icyphox:
Not sure about the test failure — doesn't appear related my change.

configx/provider.go
```diff
@@ -255,26 +252,26 @@ func (p *Provider) newKoanf() (*koanf.Koanf, error) {
...
 }
 
 func (p *Provider) traceConfig(ctx context.Context, k *koanf.Koanf, opName string) {
-	span, ctx := p.startSpan(ctx, opName)
-	defer span.Finish()
+	ctx, span := p.startSpan(ctx, opName)
+	defer span.End()
 
-	span.SetTag("component", "github.com/ory/x/configx")
+	span.SetAttributes(attribute.String("component", "github.com/ory/x/configx"))
```
Demonsthere, line 272: I think the component url should be set as a const

configx/provider.go
```diff
@@ -476,26 +477,17 @@ func (p *Provider) CORS(prefix string, defaults cors.Options) (cors.Options, boo
...
-				LocalAgentAddress: p.String("tracing.providers.jaeger.local_agent_address"),
-				MaxTagValueLength: p.IntF("tracing.providers.jaeger.max_tag_value_length", jaeger.DefaultMaxTagValueLength),
-				Propagation: stringsx.Coalesce(
-					os.Getenv("JAEGER_PROPAGATION"),
-					p.String("tracing.providers.jaeger.propagation"),
-				),
-			},
-			Zipkin: &tracing.ZipkinConfig{
-				ServerURL: p.String("tracing.providers.zipkin.server_url"),
+				LocalAgentHost: p.String("tracing.providers.jaeger.local_agent_host"),
```
Demonsthere, line 489: Splitting the address into separate variables will be a breaking change, maybe we should leave `local_agent_address` as a fallback instead of doing it as a binary change?

configx/provider_test.go
```diff
@@ -149,16 +149,15 @@ func TestAdvancedConfigs(t *testing.T) {
...
 			envs: [][2]string{
 				{"DSN", "sqlite:///var/lib/sqlite/db.sqlite?_fk=true"},
-				{"TRACING_PROVIDER", "jaeger"},
-				{"TRACING_PROVIDERS_JAEGER_SAMPLING_SERVER_URL", "http://jaeger:5778/sampling"},
-				{"TRACING_PROVIDERS_JAEGER_LOCAL_AGENT_ADDRESS", "jaeger:6832"},
-				{"TRACING_PROVIDERS_JAEGER_SAMPLING_TYPE", "const"},
-				{"TRACING_PROVIDERS_JAEGER_SAMPLING_VALUE", "1"},
+				{"OTEL_EXPORTER_JAEGER_SAMPLING_SERVER_URL", "http://jaeger:5778/sampling"},
+				{"OTEL_EXPORTER_JAEGER_AGENT_HOST", "jaeger"},
+				{"OTEL_EXPORTER_JAEGER_AGENT_PORT", "6832"},
```
Demonsthere, line 154: Please see my other comment, imho we should have support for all of the 3 values:
- JAEGER_LOCAL_AGENT_ADDRESS
- JAEGER_AGENT_HOST
- JAEGER_AGENT_PORT

otelx/config.go
```diff
@@ -0,0 +1,40 @@
+package otelx
+
+import (
+	"bytes"
+	_ "embed"
+	"io"
+)
+
+type JaegerConfig struct {
+	LocalAgentHost string         `json:"local_agent_host"`
```
Demonsthere, line 10: Please see my other comments :)

otelx/middleware.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/julienschmidt/httprouter"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func healthFilter(r *http.Request) bool {
```
Demonsthere, line 15:
```suggestion
func isHealthFilter(r *http.Request) bool {
```

otelx/middleware.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Middleware to satisfy httprouter.Handle. The URL path is used as
+// the span name.
+func WrapHTTPRouter(next httprouter.Handle) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, ps httprouter.Params) {
+		operation := r.URL.Path
+		var tracer trace.Tracer
+		if span := trace.SpanFromContext(r.Context()); span.SpanContext().IsValid() {
+			tracer = span.TracerProvider().Tracer("github.com/ory/x/otelx")
```
Demonsthere, line 45: Please extract the values as consts :)

otelx/otel.go
```diff
@@ -0,0 +1,142 @@
...
+//    OTEL_EXPORTER_JAEGER_AGENT_HOST
+//    OTEL_EXPORTER_JAEGER_AGENT_PORT
+//
+// Optionally, Config.Providers.Jaeger.LocalAgentAddress can be set.
+// NOTE: If Config.Providers.Jaeger.Sampling.ServerURL is not specfied,
+// AlwaysSample is used.
+func (t *Tracer) setup(name string) error {
+	switch t.Config.Provider {
+	case "jaeger":
+		host := stringsx.Coalesce(
```
Demonsthere, line 63: I think this logic should be moved to a seperate function so we can have:

```golang
switch t.Config.Provider {
	case "jaeger": setupJager()
	case "zipkin": setupZipkin()
```

otelx/sql/instrumentedsql.go
```diff
@@ -0,0 +1,77 @@
...
+	return span{parent: trace.SpanFromContext(ctx), tracer: t}
+}
+
+func (s span) NewChild(name string) instrumentedsql.Span {
+	if s.ctx == nil {
+		return s
+	}
+
+	var parent trace.Span
+	tp := otel.GetTracerProvider().Tracer("github.com/ory/x/otelx/sql")
```
Demonsthere, line 40: Please export to consts

popx/migrator.go
```diff
@@ -504,24 +506,22 @@ func (m *Migrator) DumpMigrationSchema(ctx context.Context) error {
...
-	defer span.Finish()
+	defer span.End()
 
 	return f(ctx, span)
 }
 
-func (m *Migrator) startSpan(ctx context.Context, opName string) (opentracing.Span, context.Context) {
-	tracer := opentracing.GlobalTracer()
+func (m *Migrator) startSpan(ctx context.Context, opName string) (trace.Span, context.Context) {
+	tracer := otel.Tracer("github.com/ory/x/popx")
```
Demonsthere, line 517: Please export to consts

Demonsthere (commented):
Hi there :) 
Adding some initial review comments, keep up the good work!

otelx/config_test.go
```diff
@@ -32,8 +32,9 @@ func TestConfigSchema(t *testing.T) {
 			Provider:    "jaeger",
 			Providers: ProvidersConfig{
 				Jaeger: JaegerConfig{
-					LocalAgentHost: "localhost",
-					LocalAgentPort: 6831,
+					LocalAgentAddress: "localhost:6831",
```
Demonsthere, line 35: For tests, we should check the following scenario:
if `LocalAgentAddress` is given, use it, else fallback to `LocalAgentHost` and `LocalAgentPort`. So that if all 3 are given, `LocalAgentAddress` takes precedence

icyphox: done!

Demonsthere:
Great! It seems all that is left to do now i so to make the tests work :D https://github.com/ory/x/runs/5991843392?check_suite_focus=true#step%3A12%3A1202=

Demonsthere (approved):
LGTM :)

configx/provider.go
```diff
@@ -284,13 +283,17 @@ func (p *Provider) traceConfig(ctx context.Context, k *koanf.Koanf, opName strin
...
-			fields = append(fields, log.Object(key, k.Get(key)))
+			// XXX: The issue here is, we can't guess what attribute type to
+			// use without trying every type supported in Koanf and checking for
+			// a zero value. Is fmt.Sprint the best way?
+			fields = append(fields, attribute.String(key, fmt.Sprint(k.Get(key))))
 		}
 	}
 
-	span.LogFields(fields...)
+	// TODO: better name for trace event?
```
zepatrik, line 295: address todo

configx/provider.go
```diff
@@ -476,26 +479,17 @@ func (p *Provider) CORS(prefix string, defaults cors.Options) (cors.Options, boo
...
+			Jaeger: otelx.JaegerConfig{
+				Sampling: otelx.JaegerSampling{
 					ServerURL: p.String("tracing.providers.jaeger.sampling.server_url"),
 				},
-				LocalAgentAddress: p.String("tracing.providers.jaeger.local_agent_address"),
-				MaxTagValueLength: p.IntF("tracing.providers.jaeger.max_tag_value_length", jaeger.DefaultMaxTagValueLength),
-				Propagation: stringsx.Coalesce(
-					os.Getenv("JAEGER_PROPAGATION"),
-					p.String("tracing.providers.jaeger.propagation"),
-				),
```
zepatrik, line 495: Where did these features go to?

configx/provider.go
```diff
@@ -476,26 +479,17 @@ func (p *Provider) CORS(prefix string, defaults cors.Options) (cors.Options, boo
...
-			},
-			Zipkin: &tracing.ZipkinConfig{
```
zepatrik, line 497: Does this mean we don't support zipkin anymore?

otelx/config_test.go
```diff
@@ -0,0 +1,76 @@
...
+		assert.NoError(t, schema.Validate(bytes.NewBufferString(rawConfig)))
+	})
+
+	t.Run("case=AddressTakesPrecedence", func(t *testing.T) {
+		conf := Config{
+			ServiceName: "Ory X",
+			Provider:    "jaeger",
+			Providers: ProvidersConfig{
+				Jaeger: JaegerConfig{
+					LocalAgentAddress: "foo:6831",
```
zepatrik, line 62: Correct me if I am wrong, but shouldn't `configx.Provider` use this field? It does not seem to use it.

otelx/jaeger.go
```diff
@@ -0,0 +1,85 @@
...
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func configureHostPort(c Config) (host, port string) {
+	address := c.Providers.Jaeger.LocalAgentAddress
+	splitAddr := strings.Split(address, ":")
```
zepatrik, line 21: This is a bit problematic with IPv6 addresses. I recommend to instead use `net.SplitHostPort`.

otelx/middleware_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/negroni"
+	"go.opentelemetry.io/otel"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+)
+
+func TestShouldNotTraceHealthEndpoint(t *testing.T) {
```
zepatrik, line 15: We should also have a test with a non-health endpoint and check that it gets traced.

otelx/otel_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	})
+	require.NoError(t, err)
+
+	trc := ot.Tracer()
+	_, span := trc.Start(context.Background(), "testSpan")
+	span.SetAttributes(attribute.Bool("testAttribute", true))
+	span.End()
+
+	<-done
```
zepatrik, line 52: This means, in case of failure, the test just runs into timeout right? That's not very nice for debugging/CI.
Easy fix:
```suggestion
	select {
		case <-done:
		case <-time.After(10*time.Second):
			t.Fail("expected to receive span, but did not receive any")
	}
```

otelx/otel_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"testing"
+
+	"github.com/ory/x/logrusx"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+func TestJaegerTracer(t *testing.T) {
+	host := "127.0.0.1"
+	port := "6831"
```
zepatrik, line 16: Maybe use `freeport.Port` or `0` to get a free port.
If you use port 0 here, you can get the actual port from the socket.

otelx/otel_test.go
```diff
@@ -0,0 +1,53 @@
...
+	done := make(chan struct{})
+	go func(addr string) {
+		t.Log("Starting test UDP server for Jaeger spans")
+		udpAddr, _ := net.ResolveUDPAddr("udp", addr)
+		srv, _ := net.ListenUDP("udp", udpAddr)
```
zepatrik, line 21:
```suggestion
		udpAddr, err := net.ResolveUDPAddr("udp", addr)
		require.NoErr(t, err)
		srv, err := net.ListenUDP("udp", udpAddr)
		require.NoErr(t, err)
```

otelx/otel_test.go
```diff
@@ -0,0 +1,53 @@
...
+		for {
+			buf := make([]byte, 2048)
+			_, conn, _ := srv.ReadFromUDP(buf)
```
zepatrik, line 24:
```suggestion
			_, conn, err := srv.ReadFromUDP(buf)
			require.NoErr(t, err)
```

otelx/otel_test.go
zepatrik: If you ignore those errors, it is literally impossible to figure out why the test is failing.

zepatrik: Ah wait, you can't use `require` in other goroutines. Use an errgroup instead and return the error.

otelx/sql/instrumentedsql.go
```diff
@@ -0,0 +1,78 @@
...
+	if s.ctx == nil {
+		return s
+	}
+
+	var parent trace.Span
+	tp := otel.GetTracerProvider().Tracer(tracingComponent)
+	// if s.parent == nil {
+	// 	_, parent = tp.Start(context.Background(), name)
+	// 	return span{parent: parent, tracer: s.tracer}
+	// } else {
```
zepatrik, line 45: remove?

proxy/proxy.go
```diff
@@ -54,7 +56,11 @@ const (
 // director is a custom internal function for altering a http.Request
 func director(o *options) func(*http.Request) {
 	return func(r *http.Request) {
-		c, err := o.hostMapper(r.Context(), r)
+		ctx := r.Context()
+		newCtx, span := otel.GetTracerProvider().Tracer("").Start(ctx, "x.proxy")
```
zepatrik, line 60: It is better to overwrite the variable, as you don't want to, by accident, use the original context ever again.
```suggestion
		ctx, span := otel.GetTracerProvider().Tracer("").Start(ctx, "x.proxy")
```

zepatrik (changes requested):
This looks pretty good already :+1: 
Can we remove the old tracing package as well?
Also, why not stick with the current `local_agent_address` config key to avoid breaking changes and all the edge cases in the first place?

configx/provider.go
icyphox: Removal of LocalAgentAddress was erroneous; Propagation is no longer configurable but I suppose it could be; MaxTagValueLength is not a thing anymore. OpenTelemetry isn't 1:1 with jaeger-client-go and OpenTracing.

configx/provider.go
icyphox: We do! It's all just simplified into one single configuration of a TextMapPropogator: https://github.com/ory/x/blob/dfef73e51c4a1de23d89b8696ae87545d82a1c39/otelx/jaeger.go#L81

otelx/otel_test.go
icyphox: This was how I'd written it initially, but sometimes, it takes a while for the span to come through (for whatever reason) and it's hard to accurately predict the duration. I've gone ahead and added a 15s limit, but this could potentially introduce flakiness.

icyphox:
> Can we remove the old tracing package as well?

Done.

> Also, why not stick with the current `local_agent_address` config key to avoid breaking changes and all the edge cases in the first place?

Also done. Not sure why that didn't occur to me 😅

otelx/config.schema.json
```diff
@@ -0,0 +1,84 @@
...
+              "allOf": [
+                {
+                  "properties": {
+                    "server_url": {
+                      "type": "string",
+                      "description": "The address of jaeger-agent's HTTP sampling server",
+                      "format": "uri"
+                    }
+                  }
+                }
```
zepatrik, line 71: why not just
```suggestion
              "properties": {
                    "server_url": {
                      "type": "string",
                      "description": "The address of jaeger-agent's HTTP sampling server",
                      "format": "uri"
                    }
                  }
                }
```

zepatrik: and we should add `additionalProperties: false`

otelx/middleware.go
```diff
@@ -0,0 +1,66 @@
...
+	return opts
+}
+
+func NewHandler(handler http.Handler, operation string) http.Handler {
+	return otelhttp.NewHandler(handler, operation, filterOpts()...)
+}
+
+// Middleware to satisfy httprouter.Handle. The URL path is used as
+// the span name.
+func WrapHTTPRouter(next httprouter.Handle) httprouter.Handle {
```
zepatrik, line 42: This works of course, but wouldn't it be nicer to implement the [negroni middleware](https://pkg.go.dev/github.com/urfave/negroni#Handler) interface so that it works independed of `httprouter`?

otelx/middleware_test.go
```diff
@@ -0,0 +1,47 @@
...
+			req := httptest.NewRequest(http.MethodGet, "https://api.example.com/"+test.path, nil)
+			h := NewHandler(negroni.New(), "test op")
+			h.ServeHTTP(negroni.NewResponseWriter(httptest.NewRecorder()), req)
+
+			spans := recorder.Ended()
+			if strings.Contains(test.path, "health") {
+				assert.Len(t, spans, 0)
+			} else {
+				assert.Len(t, spans, 1)
+			}
```
zepatrik, line 44: It would be a bit better to have a field in the testcase struct which is e.g. `expectedSpans int` to make it explicit, but this is also fine.

zepatrik (commented):
Nice, this looks very good now :+1: Just some minor improvements :wink:

otelx/config.schema.json
```diff
@@ -0,0 +1,81 @@
...
+              "description": "The address of the jaeger-agent where spans should be sent to.",
+              "oneOf": [
+                {
+                  "pattern": "^\\[(([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))]:([0-9]*)$"
+                },
+                {
+                  "pattern": "^([0-9]{1,3}\\.){3}[0-9]{1,3}:([0-9]*)$"
+                },
+                {
+                  "format": "uri"
```
zepatrik, line 48: Format `uri` requires a scheme, but the code does not allow for parsing schemes, only `hostname:port` combinations.

otelx/config.schema.json
```diff
@@ -0,0 +1,81 @@
...
+                },
+                "additionalProperties": false
+              },
+              "examples": [
+                {
+                  "server_url": "http://localhost:5778/sampling"
+                }
+              ]
+            }
```
zepatrik, line 75:
```suggestion
                  "format": "uri",
                  "examples": [
                    "http://localhost:5778/sampling"
                  ]
                },
                "additionalProperties": false
              }
            }
```

otelx/config.schema.json
```diff
@@ -0,0 +1,81 @@
...
+                  "server_url"
+                ]
+              },
+              "properties": {
+                "server_url": {
+                  "type": "string",
+                  "description": "The address of jaeger-agent's HTTP sampling server",
+                  "format": "uri"
+                },
+                "additionalProperties": false
```
zepatrik, line 68:
```suggestion
              "additionalProperties": false,
              "properties": {
                "server_url": {
                  "type": "string",
                  "description": "The address of jaeger-agent's HTTP sampling server",
                  "format": "uri"
                }
```

otelx/otel.go
```diff
@@ -0,0 +1,65 @@
...
+// Creates a new no-op tracer.
+func NewNoop(l *logrusx.Logger, c *Config) *Tracer {
+	tp := trace.NewNoopTracerProvider()
+	t := &Tracer{Config: c, l: l, tracer: tp.Tracer("")}
+	return t
+}
+
+// setup sets up the tracer.
+func (t *Tracer) setup(name string) error {
+	switch t.Config.Provider {
```
zepatrik, line 37: You can use https://github.com/ory/x/blob/e49f326710518506477a020f5d7eeef02f92de5c/stringsx/switch_case_test.go#L59-L66
That will automatically create a nice error with all valid cases.

popx/migrator.go
```diff
@@ -85,8 +85,11 @@ func (m *Migrator) Up(ctx context.Context) error {
 // If step <= 0 all pending migrations are run.
 func (m *Migrator) UpTo(ctx context.Context, step int) (applied int, err error) {
 	span, ctx := m.startSpan(ctx, MigrationUpOpName)
-	defer span.Finish()
-	span.LogFields(log.Int("up_to_step", step))
+	defer span.End()
+	// span.LogFields(log.Int("up_to_step", step))
```
zepatrik, line 89: remove?

zepatrik (changes requested):
Just very minor improvements and we're good to go :tada:

otelx/middleware.go
icyphox: I just realized that otelx.WrapHTTPRouter isn't really required since the actual http.Handler above is already being wrapped. Our tracing works just fine without this.

otelx/otel.go
icyphox: Ok I've added this. I hope I've used it right. :)

otelx/config.schema.json
icyphox: Yeah I get that, but removing `"format": "urI"` causes tests to fail and I'm not sure why.

zepatrik (approved):

aeneasr:
Guys, please don't remove packages just because a new package has been introduced. This makes it impossible to upgrade ory/x without changing all dependent code of the stuff that has been removed.

PR #480, opened 2022-03-11
chore: remove all dependencies to older ory/x versions

zepatrik:
This should greatly reduce the number of dependencies, as ory/x should now not have any more dependency on older versions of itself.

```
$ go mod graph | cut -d" " -f2 | grep ory/x
$ git checkout v0.0.355
$ go mod graph | cut -d" " -f2 | grep ory/x
github.com/ory/x@v0.0.250
github.com/ory/x@v0.0.288
github.com/ory/x@v0.0.272
github.com/ory/x@v0.0.110
github.com/ory/x@v0.0.205
github.com/ory/x@v0.0.205
github.com/ory/x@v0.0.84
github.com/ory/x@v0.0.93
github.com/ory/x@v0.0.127
```

I'd say this is a success?


zepatrik:
OK this ACTUALLY worked :tada: 

Keto dependencies dropped from a whopping `11235` to `7328` :sunglasses:

PR #484, opened 2022-03-24
fix: add ability to render to Ory Kratos config YAML

aeneasr:
@vinckr @tomekpapiernik this patch will allow the CLI to render ory cloud configs to ory kratos open source configs. it's pretty easy, just use `--format kratos-config` yaml instead of e.g. `--format json`:

```
$ ory get project <id> --format kratos-config
You are authenticated as: dev+orycye2eda2f162daf6142dd.r8hxnywu0y7t0mre@ory.dev
cookies:
  domain: strange-antonelli-rmss6482s9.projects.oryapis.com
  path: /
  same_site: Lax
courier:
  smtp:
    from_name: ory-cy-e2e-da2f162d-af61-42dd-90dc-e3fcfa7c84a0-Updated Project from
      JSON via Ory
  templates:
    recovery:
```

This also works with patch/update :)

In Ory Kratos, the difficulty will probably be to load both your "server" config (e.g. ports) and the cloud config and merge them. I **think** it should be possible with:

```
$ kratos serve -c the-config-from-cloud.yml -c your-local-overrides.yml
```

in that order (overrides comes last). I'm not 100% sure how well this will work but maybe you could try this and write a guide for it?

The local-overrides would contain things like:

```
dsn: conneciton-to-database
secrets:
  default:
    - ...
server:
  public:
    base_url: ...
    port: ...
...
```

We don't have to give a full guide on how to configure everything, but it would probably be good to have a basic working example. Please keep in mind that the "merge" strategy for the configs is pretty basic. It will not be possible to merge things like arrays or deeply nested keys. So yeah, it's only possible to merge these things up to a certain level. It won't be possible, for example, to configure parts of a social sign in provider in your overrides file AND merge them with your cloud config.

See also:

- https://github.com/knadh/koanf/issues/125
- https://github.com/knadh/koanf/pull/126


zepatrik (commented):
I think it would be better to not mix format and the object you want to get in the end. It will make it more complex to output more stuff in the future, as it implicitly has e.g. yaml or JSON output format, but the user can't control that.
Imagine you build some automation and want to export the current Kratos config of your prod system to then modify with `jq` and then spin up a local test instance with the same config. Sure, in this specific case you could also use `yq` or add a second config file.
Alternatively, if you can e.g. also export the list of collaborators later on, is that JSON or yaml?
With `kubectl` you have the resource type as an argument, so `ory get project <id>` and `ory get kratos-config <project-id>` or `ory get kratos-config --project <id>` makes more sense IMO.
We can easily add more resource types as we need, and the format still does what you expect it to do: control the output format, not content.

aeneasr:
One problem I saw here was that you need two calls to patch & render:

```
ory patch ...
ory get kratos-config ...
```

which would make the docs a bit ugly. I mean there's another option which would be to pipe it:

```
ory get project --format json | ory render kratos-config
ory import kratos-config path/to/file | ory put project -f -
```

but this doesn't really work well on windows and is a bit difficult for newcomers (who in their right mind can remember this sequence of commands?).

Another option would be to introduce a `--transform` property that does this. Then again, it doesn't really make sense to export the kratos config as a CSV table...

zepatrik:
Still I think format should control how the output looks, not what you see...

But why not stick with the resource types and have
```
ory patch project ...
ory patch kratos-config ...
ory patch identity-config ...
```
and then have `kratos-config` as an alias for `identity-config`.
Would also allow to copy `kubectl apply`
```
cat kratos-config.yml
> kind: kratos-config
> apiVersion: ...
> ...
ory apply --file kratos-config.yml
```

aeneasr:
Ah yes, this is kind of the approach I originally thought of in the RFC. I think that using `get` with explicit naming such as `kratos-config` is a much smarter approach than what I proposed here. I'll refactor it accordingly

aeneasr:
i triggered the release, target is 0.1.28

PR #493, opened 2022-04-21
fix: force user to relogin on session expiration closes #2342

gen1us2k:
## Checklist

      and signed the CLA.
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/cloudx/cmd_auth_test.go
+++ b/cloudx/cmd_auth_test.go
@@ -99,7 +99,7 @@ func TestAuthenticator(t *testing.T) {
 			changeAccessToken(t, configDir)
 			_, stderr, err := cmd.ExecDebug(t, nil, "list", "projects", "-q")
 			require.Error(t, err)
-			assert.Equal(t, "your session expired and you can not reauthenticate when flag --quiet is set", err.Error())
+			assert.Equal(t, "Your session has expired and you cannot reauthenticate when the --quiet flag is set", err.Error())
 			assert.NotContains(t, stderr, "Your CLI session has expired. Do you wish to log in again as")
 		})
 
--- a/cloudx/cmd_auth_test.go
+++ b/cloudx/cmd_auth_test.go
@@ -75,10 +75,10 @@ func TestAuthenticator(t *testing.T) {
 			expectSignInSuccess(t)
 			changeAccessToken(t, configDir)
 			var r bytes.Buffer
-			r.WriteString("n\n") // Your CLI session has expired. Do you wish to login again with <email>?
+			r.WriteString("n\n") // Your CLI session has expired. Do you wish to login again as <email>?
 			_, stderr, err := cmd.ExecDebug(t, &r, "list", "projects")
 			require.Error(t, err)
-			assert.Contains(t, stderr, "Your CLI session has expired. Do you wish to log in again with")
+			assert.Contains(t, stderr, "Your CLI session has expired. Do you wish to log in again as")
 		})
 
 		t.Run("set up 2fa", func(t *testing.T) {
--- a/cloudx/cmd_auth_test.go
+++ b/cloudx/cmd_auth_test.go
@@ -70,6 +70,17 @@ func TestAuthenticator(t *testing.T) {
 			expectSignInSuccess(t)
 		})
 
+		t.Run("forced to reauthenticate on session expiration", func(t *testing.T) {
+			cmd := configAwareCmd(configDir)
+			expectSignInSuccess(t)
+			changeAccessToken(t, configDir)
+			var r bytes.Buffer
+			r.WriteString("n\n") // Your CLI session has expired. Do you wish to login again with <email>?
+			_, stderr, err := cmd.ExecDebug(t, &r, "list", "projects")
+			require.Error(t, err)
+			assert.Contains(t, stderr, "Your CLI session has expired. Do you wish to log in again with")
+		})
+
 		t.Run("set up 2fa", func(t *testing.T) {
 			expectSignInSuccess(t)
 			ac := readConfig(t, configDir)
--- a/cloudx/cmd_auth_test.go
+++ b/cloudx/cmd_auth_test.go
@@ -81,6 +81,28 @@ func TestAuthenticator(t *testing.T) {
 			assert.Contains(t, stderr, "Your CLI session has expired. Do you wish to log in again as")
 		})
 
+		t.Run("user is able to reauthenticate on session expiration", func(t *testing.T) {
+			cmd := configAwareCmd(configDir)
+			expectSignInSuccess(t)
+			changeAccessToken(t, configDir)
+			var r bytes.Buffer
+			r.WriteString("y\n") // Your CLI session has expired. Do you wish to login again as <email>?
+			_, stderr, err := cmd.ExecDebug(t, &r, "list", "projects")
+			require.Error(t, err)
+			assert.Contains(t, stderr, "Your CLI session has expired. Do you wish to log in again as")
+			expectSignInSuccess(t)
+		})
+
+		t.Run("expired session with quiet flag returns error", func(t *testing.T) {
+			cmd := configAwareCmd(configDir)
+			expectSignInSuccess(t)
+			changeAccessToken(t, configDir)
+			_, stderr, err := cmd.ExecDebug(t, nil, "list", "projects", "-q")
+			require.Error(t, err)
+			assert.Equal(t, "your session expired and you can not reauthenticate when flag --quiet is set", err.Error())
+			assert.NotContains(t, stderr, "Your CLI session has expired. Do you wish to log in again as")
+		})
+
 		t.Run("set up 2fa", func(t *testing.T) {
 			expectSignInSuccess(t)
 			ac := readConfig(t, configDir)
--- a/cloudx/handler.go
+++ b/cloudx/handler.go
@@ -238,7 +238,7 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
 		_, resp, err := client.V0alpha2Api.ToSession(h.ctx).XSessionToken(c.SessionToken).Execute()
 		if resp.StatusCode == http.StatusUnauthorized || err != nil {
 			if h.isQuiet {
-				return nil, errors.New("can not sign in or sign up when flag --quiet is set")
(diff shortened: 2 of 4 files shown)


aeneasr (changes requested):
Could you please add a test for this? :)

cloudx/handler.go
```diff
@@ -232,6 +232,27 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
 
 	if len(c.SessionToken) > 0 {
 		_, _ = fmt.Fprintf(h.verboseErrWriter, "You are authenticated as: %s\n", c.IdentityTraits.Email)
+		client, err := newKratosClient()
+		if err != nil {
+			return nil, err
+		}
+		_, resp, err := client.V0alpha2Api.ToSession(h.ctx).XSessionToken(c.SessionToken).Execute()
+		if resp.StatusCode == http.StatusUnauthorized || err != nil {
+			ok, err := cmdx.AskScannerForConfirmation(fmt.Sprintf("Your CLI session has expired. Do you wish to log in again with \"%s\"?", c.IdentityTraits.Email), h.stdin, h.verboseErrWriter)
```
zepatrik, line 241:
```suggestion
			ok, err := cmdx.AskScannerForConfirmation(fmt.Sprintf("Your CLI session has expired. Do you wish to log in again as \"%s\"?", c.IdentityTraits.Email), h.stdin, h.verboseErrWriter)
```

cloudx/handler.go
```diff
@@ -232,6 +232,27 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
 
 	if len(c.SessionToken) > 0 {
 		_, _ = fmt.Fprintf(h.verboseErrWriter, "You are authenticated as: %s\n", c.IdentityTraits.Email)
```
zepatrik, line 234: Potentially move this to after the check? Would make more sense form a UX perspective.

cloudx/handler.go
```diff
@@ -232,6 +232,27 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
...
+			return nil, err
+		}
+		_, resp, err := client.V0alpha2Api.ToSession(h.ctx).XSessionToken(c.SessionToken).Execute()
+		if resp.StatusCode == http.StatusUnauthorized || err != nil {
+			ok, err := cmdx.AskScannerForConfirmation(fmt.Sprintf("Your CLI session has expired. Do you wish to log in again with \"%s\"?", c.IdentityTraits.Email), h.stdin, h.verboseErrWriter)
+			if err != nil {
+				return nil, err
+			}
+			if ok {
+				if err := h.SignOut(); err != nil {
```
zepatrik, line 246: Does this only delete the current session but not the email address?

cloudx/handler.go
```diff
@@ -232,6 +232,27 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
...
+					return nil, err
+				}
+				c, err := h.Authenticate()
```
zepatrik, line 249: And will then here the email address be pre-filled?

zepatrik (changes requested):

cloudx/handler.go
gen1us2k: It removes everything and creates `AuthContext` with null values

aeneasr:
Tests are currently failing - please take a look :)

gen1us2k:
@aeneasr Fixed tests.

cloudx/handler.go
```diff
@@ -480,14 +500,14 @@ func (h *SnakeCharmer) ListProjects() ([]cloud.Project, error) {
 	if err != nil {
 		return nil, err
 	}
-
```
aeneasr, line 483: Can you please revert these newline changes? They make the code less readable. We try to separate blocks of logic using a singular newline :)

cloudx/handler.go
```diff
@@ -231,6 +231,27 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
 	}
 
 	if len(c.SessionToken) > 0 {
+		client, err := newKratosClient()
+		if err != nil {
+			return nil, err
+		}
+		_, resp, err := client.V0alpha2Api.ToSession(h.ctx).XSessionToken(c.SessionToken).Execute()
+		if resp.StatusCode == http.StatusUnauthorized || err != nil {
+			ok, err := cmdx.AskScannerForConfirmation(fmt.Sprintf("Your CLI session has expired. Do you wish to log in again as \"%s\"?", c.IdentityTraits.Email), h.stdin, h.verboseErrWriter)
```
aeneasr, line 240: If the quiet flag is set, this should error and not prompt!

cloudx/handler.go
aeneasr: yes

aeneasr (changes requested):

cloudx/cmd_auth_test.go
```diff
@@ -70,6 +70,17 @@ func TestAuthenticator(t *testing.T) {
...
 
+		t.Run("forced to reauthenticate on session expiration", func(t *testing.T) {
+			cmd := configAwareCmd(configDir)
+			expectSignInSuccess(t)
+			changeAccessToken(t, configDir)
+			var r bytes.Buffer
+			r.WriteString("n\n") // Your CLI session has expired. Do you wish to login again as <email>?
+			_, stderr, err := cmd.ExecDebug(t, &r, "list", "projects")
+			require.Error(t, err)
+			assert.Contains(t, stderr, "Your CLI session has expired. Do you wish to log in again as")
```
aeneasr, line 81: Can you please add the log in sequence here (see the other tests), to ensure that we are actually able to log in properly?

cloudx/handler.go
```diff
@@ -231,6 +231,30 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
...
 
 	if len(c.SessionToken) > 0 {
+		client, err := newKratosClient()
+		if err != nil {
+			return nil, err
+		}
+		_, resp, err := client.V0alpha2Api.ToSession(h.ctx).XSessionToken(c.SessionToken).Execute()
+		if resp.StatusCode == http.StatusUnauthorized || err != nil {
+			if h.isQuiet {
+				return nil, errors.New("can not sign in or sign up when flag --quiet is set")
```
aeneasr, line 241: Can you please add an error message that explicitly states that the session has expired? Please also add a test for this behaviour.

cloudx/handler.go
```diff
@@ -231,6 +231,30 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
 	}
 
 	if len(c.SessionToken) > 0 {
+		client, err := newKratosClient()
+		if err != nil {
+			return nil, err
+		}
+		_, resp, err := client.V0alpha2Api.ToSession(h.ctx).XSessionToken(c.SessionToken).Execute()
+		if resp.StatusCode == http.StatusUnauthorized || err != nil {
```
aeneasr, line 239: If `err != nil`, then this will panic because `resp` can be `nil`.

cloudx/handler.go
```diff
@@ -231,6 +231,30 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
...
+			if h.isQuiet {
+				return nil, errors.New("can not sign in or sign up when flag --quiet is set")
+			}
+			ok, err := cmdx.AskScannerForConfirmation(fmt.Sprintf("Your CLI session has expired. Do you wish to log in again as \"%s\"?", c.IdentityTraits.Email), h.stdin, h.verboseErrWriter)
+			if err != nil {
+				return nil, err
+			}
+			if ok {
```
aeneasr, line 247: What happens if `ok` is `false`? It looks like we now just ignore the error. Please fix this and add a test for this behavior!

cloudx/handler_test.go
```diff
@@ -103,6 +103,16 @@ func configPasswordAwareCmd(configDir, password string) *cmdx.CommandExecuter {
...
 }
 
+func changeAccessToken(t *testing.T, configDir string) {
+	ac := readConfig(t, configDir)
+	ac.SessionToken = "12341234"
+	data, err := json.Marshal(ac)
+	require.NoError(t, err)
+	err = os.WriteFile(configDir, data, 0644)
+	require.NoError(t, err)
+
```
aeneasr, line 113:
```suggestion
	require.NoError(t, os.WriteFile(configDir, data, 0644))
```

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good and I have some ideas how to improve it further :)

cloudx/handler.go
gen1us2k: I handled this case

cloudx/handler.go
gen1us2k: I think that checking the session for the `nil` value would be better in that case.

cloudx/handler.go
```diff
@@ -231,6 +231,31 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
...
 
 	if len(c.SessionToken) > 0 {
+		client, err := newKratosClient()
+		if err != nil {
+			return nil, err
+		}
+		sess, _, err := client.V0alpha2Api.ToSession(h.ctx).XSessionToken(c.SessionToken).Execute()
+		if sess == nil || err != nil {
+			if h.isQuiet {
+				return nil, errors.New("your session expired and you can not reauthenticate when flag --quiet is set")
```
icyphox, line 241:
```suggestion
				return nil, errors.New("Your session has expired and you cannot reauthenticate when the --quiet flag is set")
```

icyphox: Small nit. :)

cloudx/handler.go
```diff
@@ -231,6 +231,31 @@ func (h *SnakeCharmer) EnsureContext() (*AuthContext, error) {
...
+				if err := h.SignOut(); err != nil {
+					return nil, err
+				}
+				c, err := h.Authenticate()
+				if err != nil {
+					return nil, err
+				}
+				return c, nil
+			}
+			return nil, errors.New("your session expired")
```
icyphox, line 257: Just for consistency:
```suggestion
			return nil, errors.New("Your session has expired")
```

aeneasr (approved):

PR #501, opened 2022-05-04
feat: improve usability of `snapshotx`

zepatrik:
Currently you can either ignore all matching OR exact keys. With this PR we have the ability to use both options at the same time.

Main use case: ignore all `created_at` timestamps and only ignore the root `id`.


aeneasr (approved):
Much better .... :)

PR #502, opened 2022-05-05
feat: TLS certs auto-reload

StarAurryon:
Kubernetes is able to auto-renew certificates with cert-manager and update files in the container. It would be cool if Ory Hydra /Kratos could support auto certificate update on file change.

---

As discussed in Ory Hydra / Kratos, core components should be merged in X repo.

### DIFF
--- a/tlsx/cert_provider.go
+++ b/tlsx/cert_provider.go
@@ -3,7 +3,6 @@ package tlsx
 import (
 	"context"
 	"crypto/tls"
-	"path/filepath"
 	"sync"
 
 	"github.com/ory/x/watcherx"
@@ -28,6 +27,7 @@ type (
 
 		watchersCancel []func()
 		fsEvent        watcherx.EventChannel
+		fsEvent2       watcherx.EventChannel
 		watchersLck    sync.Mutex
 
 		ev EventChannel
@@ -60,9 +60,7 @@ func (e *ChangeEvent) String() string {
 func NewProvider(ctx context.Context, ev EventChannel) Provider {
 	p := &provider{
 		ctx: ctx,
-
-		fsEvent: make(watcherx.EventChannel, 1),
-		ev:      ev,
+		ev:  ev,
 	}
 
 	go p.watchCertificatesChanges()
@@ -152,39 +150,23 @@ func (p *provider) setWatcher(certPath, keyPath string) error {
 
 	p.deleteWatchersNoLock()
 
-	root := filepath.Dir(certPath)
-	if root == filepath.Dir(keyPath) {
-		if err := p.addDirectoryWatcher(root); err != nil && p.ev != nil {
-			return err
-		}
-	} else {
-		if err := p.addFileWatcher(certPath); err != nil && p.ev != nil {
-			return err
-		}
-
-		if err := p.addFileWatcher(keyPath); err != nil && p.ev != nil {
-			return err
-		}
-	}
-	return nil
-}
-
-func (p *provider) addDirectoryWatcher(fsPath string) error {
-	ctx, cancel := context.WithCancel(p.ctx)
-	_, err := watcherx.WatchDirectory(ctx, fsPath, p.fsEvent)
-	if err != nil {
-		cancel()
+	if err := p.addFileWatcher(certPath, &p.fsEvent); err != nil && p.ev != nil {
 		return err
 	}
 
-	p.watchersCancel = append(p.watchersCancel, cancel)
+	if err := p.addFileWatcher(keyPath, &p.fsEvent2); err != nil && p.ev != nil {
+		return err
+	}
 
 	return nil
 }
 
-func (p *provider) addFileWatcher(fsPath string) error {
+func (p *provider) addFileWatcher(fsPath string, fsEvent *watcherx.EventChannel) error {
 	ctx, cancel := context.WithCancel(p.ctx)
-	_, err := watcherx.WatchFile(ctx, fsPath, p.fsEvent)
+
+	*fsEvent = make(watcherx.EventChannel, 1)
+
+	_, err := watcherx.WatchFile(ctx, fsPath, *fsEvent)
 	if err != nil {
 		cancel()
 		return err
--- a/tlsx/cert_provider.go
+++ b/tlsx/cert_provider.go
@@ -3,7 +3,6 @@ package tlsx
 import (
 	"context"
 	"crypto/tls"
-	"path"
 	"sync"
 	"time"
 
@@ -114,15 +113,8 @@ func (p *provider) setWatcher(certPath, keyPath string) {
 
 	p.deleteWatchersNoLock()
 
-	certPath = path.Dir(certPath)
-	keyPath = path.Dir(keyPath)
-
 	if err := p.addWatcher(certPath); err != nil {
 		p.logger.WithError(err).Fatalf("Could not create watcher with path: " + certPath)
-	}
-
-	if certPath == keyPath {
-		return
 	}
 
 	if err := p.addWatcher(keyPath); err != nil {
@@ -133,7 +125,7 @@ func (p *provider) setWatcher(certPath, keyPath string) {
 
 func (p *provider) addWatcher(fsPath string) error {
 	ctx, cancel := context.WithCancel(p.ctx)
-	_, err := watcherx.WatchDirectory(ctx, fsPath, p.fsEvent)
+	_, err := watcherx.WatchFile(ctx, fsPath, p.fsEvent)
 	if err != nil {
 		cancel()
 		return err
@@ -164,9 +156,14 @@ func (p *provider) watchCertificatesChanges() {
 			select {
 			case <-p.ctx.Done():
 				return
-			case _, ok := <-p.fsEvent:
+			case e, ok := <-p.fsEvent:
 				if !ok {
 					return
+				}
+
+				if err, isErr := e.(*watcherx.ErrorEvent); isErr {
+					p.logger.WithError(err).Warningf("Error watching: %s", e.Source())
+					continue
 				}
 
 				p.waitForAllFilesChanges()
--- a/tlsx/cert_provider.go
+++ b/tlsx/cert_provider.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/tls"
 	"sync"
-	"time"
 
 	"github.com/ory/x/logrusx"
 	"github.com/ory/x/watcherx"
@@ -166,32 +165,14 @@ func (p *provider) watchCertificatesChanges() {
 					continue
 				}
 
-				p.waitForAllFilesChanges()
-
-				p.logger.Infof("TLS certificates changed, updating")
-				if err := p.LoadCertificates("", "", p.certPath, p.keyPath); err != nil {
-					p.logger.WithError(err).Errorf("Error in the new tls certificates")
-					return
+				if _, isChange := e.(*watcherx.ChangeEvent); isChange {
+					p.logger.Infof("TLS certificates changed, updating")
+					if err := p.LoadCertificates("", "", p.certPath, p.keyPath); err != nil {
+						p.logger.WithError(err).Errorf("Error in the new tls certificates")
+						return
+					}
 				}
 			}
 		}
 	}()
-}
-
-func (p *provider) waitForAllFilesChanges() {
-	flushUntil := time.After(2 * time.Second)
-	p.logger.Infof("TLS certificates files changed, waiting for changes to finish")
-	stop := false
-	for {
-		select {
-		case <-flushUntil:
-			stop = true
-		case <-p.fsEvent:
-			continue
-		}
-
-		if stop {
-			break
-		}
-	}
 }
--- a/tlsx/cert_provider.go
+++ b/tlsx/cert_provider.go
@@ -109,7 +109,7 @@ func (p *provider) LoadCertificates(
 	if fromFiles {
 		p.certPath = certPath
 		p.keyPath = keyPath
-		p.setWatcher(certPath, keyPath)
+		return p.setWatcher(certPath, keyPath)
 	} else {
 		p.certPath = ""
 		p.keyPath = ""
@@ -146,7 +146,7 @@ func (p *provider) setCertificates(crts []tls.Certificate) {
 	p.crtsLck.Unlock()
 }
 
-func (p *provider) setWatcher(certPath, keyPath string) {
+func (p *provider) setWatcher(certPath, keyPath string) error {
 	p.watchersLck.Lock()
 	defer p.watchersLck.Unlock()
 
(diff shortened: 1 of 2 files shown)


alnr:
Hi! I've also started to work on this feaure: see #507. The problem with using fsnotify is that it doesn't work in many circumstances where the watched file is mounted over a network in some manner. For example: Kubernetes secrets mounted as files into the pod.

My approach just reloads the cert periodically, with a configurable interval. What do you think?

StarAurryon:
Wow. Your use case is weird. Why having certs on a shared drive with K8S ?

If I take best practices, you should use cert manager, hashicorp vault etc that relies on kubernetes secrets.

If you want to stick with this use case the best thing would be to allow full config reload like with "nginx -S" if I'm not wrong.

StarAurryon:
My apologies I read half of it.

"Kubernetes secrets mounted as files into the pod." -> You should mount them as folder.

Kubernetes creates symlinks and updates them and my code is covering this use case. Therefore you are not watching the files but the folder where the files are located.

E.g.
```
/etc/nginx/ssl# ls -lra
total 4
lrwxrwxrwx 1 root root   14 Mar  5 14:00 tls.key -> ..data/tls.key
lrwxrwxrwx 1 root root   14 Mar  5 14:00 tls.crt -> ..data/tls.crt
lrwxrwxrwx 1 root root   13 Mar  5 14:00 ca.crt -> ..data/ca.crt
lrwxrwxrwx 1 root root   32 Apr 30 19:11 ..data -> ..2022_04_30_17_11_27.3071490089
drwxr-xr-x 2 root root  100 Apr 30 19:11 ..2022_04_30_17_11_27.3071490089
drwxr-xr-x 1 root root 4096 Mar  5 14:00 ..
drwxrwxrwt 3 root root  140 Apr 30 19:11 .
```

Kubernetes wil create a new folder with the date and update the .data symlink.

alnr:
Sorry, I was unclear. I am thinking of the same use case as you are (certmanager or Vault, k8s secrets).

The fsnotify approach won't work in this scenario. The cert will be renewed on disk, but fsnotify won't notify you of the change.

StarAurryon:
So I believe your issue is mounting the certs as files with subPath options:
```
  volumeMounts:
  - mountPath: /srv/kratos/tls/tls.crt
    name: certs
    subPath: tls.crt
  - mountPath: /srv/kratos/tls/tls.key
    name: certs
    subPath: tls.key
```

And this should be:
```
  volumeMounts:
  - mountPath: /srv/kratos/tls
    name: certs
```

alnr:
> And then you should watch /srv/kratos/tls with inotify, not /srv/kratos/tls/tls.crt and /srv/kratos/tls/tls.key

Aha! Smart! Have you actually confirmed you get the fsnotify notification in a live k8s cluster?

StarAurryon:
Yes working 100%. We have this on all our nginx static servers with a script based on [https://github.com/rosskukulinski/nginx-kubernetes-reload](https://github.com/rosskukulinski/nginx-kubernetes-reload).

Therefore, there is no need to discuss with the kube-api-server

httpx/cert_reload.go
```diff
@@ -0,0 +1,159 @@
...
+	CertLocation struct {
+		CertPath string
+		KeyPath  string
+	}
+
+	CertLoadFunc func() ([]tls.Certificate, error)
+
+	CertProvider struct {
+		certs   []tls.Certificate
+		mu      sync.Mutex
```
alnr, line 25: Maybe premature optimization, but this should probably be a sync/Atomic or at least an RWMutex.

httpx/cert_reload.go
```diff
@@ -0,0 +1,159 @@
...
+
+		if stop {
+			break
+		}
+	}
+}
+
+func (p *CertProvider) GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
```
alnr, line 141: If you want to keep this mutex as opposed to switching to sync/atomic, I would propose this to reduce contention:

```suggestion
	certs := p.certs
	p.mu.Unlock()
```

and use `certs` instead of `p.certs` in the following.

alnr, line 141: If you want to keep this mutex as opposed to switching to sync/atomic, I would propose this to reduce contention:

```suggestion
	certs := p.certs
	p.mu.Unlock()
```

and use `certs` instead of `p.certs` in the following.

httpx/cert_reload.go
```diff
@@ -0,0 +1,159 @@
...
+
+		certLoad     CertLoadFunc
+		certLocation CertLocation
+		logger       *logrusx.Logger
+	}
+)
+
+//NewCertProvider creates a tls.Certificate provider that reloads certificates on fs change
+//certs must not be empty
+func NewCertProvider(certs []tls.Certificate, certLocation CertLocation, certLoad CertLoadFunc, l *logrusx.Logger) *CertProvider {
```
alnr, line 36: Personally, I find this interface a little bit confusing. We give a filesystem location to the function to watch, but we never actually read from the file (instead deferring to the `certLoad` callback). Do we need this complexity? Same with the initial `certs` argument. I think the interface would be better if it just accepted the location and did all reading internally.

alnr, line 36: Personally, I find this interface a little bit confusing. We give a filesystem location to the function to watch, but we never actually read from the file (instead deferring to the `certLoad` callback). Do we need this complexity? Same with the initial `certs` argument. I think the interface would be better if it just accepted the location and did all reading internally.

httpx/cert_reload.go
```diff
@@ -0,0 +1,159 @@
...
+	ret := &CertProvider{
+		certLoad:     certLoad,
+		certLocation: certLocation,
+		logger:       l,
+	}
+	ret.load(certs)
+	ret.watchCertificatesChanges()
+
+	runtime.SetFinalizer(ret, func(ret *CertProvider) { ret.Stop() })
```
alnr, line 45: Neat!

alnr, line 45: Neat!

alnr (commented):
Cool approach. You might want to consider configuring your editor to insert a space after `//` to introduce a comment.

httpx/cert_reload.go
StarAurryon: In the same way in the comment above. We should implement a procedure in Hydra/Kratos to stop the watcher when shutting down the app if we don't want "runtime.SetFinalizer".

httpx/cert_reload.go
alnr: I'm not familiar with the Kratos and Hydro codebases, so I can't comment really 😁 . Have you taken a look at #507 though? It adds this functionality to `x/tlsx` instead of `x/httpx`.

StarAurryon: No worry I can move it to tlsx if you wish but that does not answer my question as this will be used for Hydra and Kratos then.

@aeneasr requested me to merge the common things here in x. The question still remains, should we rewrite TLS confi loading in a unified way for Kratos and Hydra ?

httpx/cert_reload.go
```diff
@@ -0,0 +1,160 @@
...
+	go func() {
+		p.logger.Infof("Starting tls certificate auto-refresh")
+		for {
+			select {
+			case _, ok := <-p.watcher.Events:
+				if !ok {
+					return
+				}
+
+				p.waitForAllFilesChanges()
```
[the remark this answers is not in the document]
StarAurryon: No the issue is more related to how kubernetes manage changes in ConfigMap / Secrets when mounted in a pod.

Basically it creates a new forlder with the new content and then update a symlink at the root. So multiple Inotify events will occur if we watch the folder. So we need to discard next changes and avoid reloading 3 times in a second.

As described above watching files does not work in k8s as they are symlink to the symlink pointing to the folder.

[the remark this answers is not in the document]
StarAurryon: In the same way I am planning to use ory/x/watcherx instead I didn't pay attention that it existed.

StarAurryon:
@aeneasr I just added a new version that should suit more to your tastes.

The remaining point is how should we handle in Hydra :
```
func GetOrCreateTLSCertificate(cmd *cobra.Command, d driver.Registry, iface config.ServeInterface) []tls.Certificate
```
Maybe this should also be migrated to tlsx. ;-)

StarAurryon:
I might be wrong but watcherx.WatchFile does not seem to work with kubernetes. Here is my test and it times out meaning no event at all.

```
	t.Run("case=notifies about symlink change in kubernetes configMap - secrets", func(t *testing.T) {
		if runtime.GOOS != "linux" {
			t.Skip("skipping test because watching symlinks on windows and macOS is not working properly")
		}

		ctx, c, dir, cancel := setup(t)
		defer cancel()

		dirOne := filepath.Join(dir, "dirOne")
		require.NoError(t, os.Mkdir(dirOne, os.ModePerm))
		fileOne := filepath.Join(dirOne, "fileOne")
		f1, err := os.Create(fileOne)
		require.NoError(t, err)
		require.NoError(t, f1.Close())

		linkDirName := filepath.Join(dir, ".data")
		require.NoError(t, os.Symlink(dirOne, linkDirName))

		linkSrcFileName := filepath.Join(linkDirName, "fileOne")
		linkFileName := filepath.Join(dir, "fileOne")
		require.NoError(t, os.Symlink(linkSrcFileName, linkFileName))

		_, err = WatchFile(ctx, linkFileName, c)
		require.NoError(t, err)

		dirTwo := filepath.Join(dir, "dirTwo")
		require.NoError(t, os.Mkdir(dirTwo, os.ModePerm))
		fileOneBis := filepath.Join(dirTwo, "fileOne")
		f1bis, err := os.Create(fileOneBis)
		require.NoError(t, err)
		_, err = fmt.Fprintf(f1bis, "New file")
		require.NoError(t, err)
		require.NoError(t, f1bis.Close())

		require.NoError(t, os.Remove(linkDirName))
		require.NoError(t, os.Symlink(dirTwo, linkDirName))
		assertRemove(t, <-c, linkFileName)
	})
```

StarAurryon:
> @aeneasr I might be wrong but watcherx.WatchFile does not seem to work with kubernetes. Here is my test and it times out meaning no event at all.
> 
> ```
> 	t.Run("case=notifies about symlink change in kubernetes configMap - secrets", func(t *testing.T) {
> 		if runtime.GOOS != "linux" {
> 			t.Skip("skipping test because watching symlinks on windows and macOS is not working properly")
> 		}
> 
> 		ctx, c, dir, cancel := setup(t)
> 		defer cancel()
> 
> 		dirOne := filepath.Join(dir, "dirOne")
> 		require.NoError(t, os.Mkdir(dirOne, os.ModePerm))
> 		fileOne := filepath.Join(dirOne, "fileOne")
> 		f1, err := os.Create(fileOne)
> 		require.NoError(t, err)
> 		require.NoError(t, f1.Close())
> 
> 		linkDirName := filepath.Join(dir, ".data")
> 		require.NoError(t, os.Symlink(dirOne, linkDirName))
> 
> 		linkSrcFileName := filepath.Join(linkDirName, "fileOne")
> 		linkFileName := filepath.Join(dir, "fileOne")
> 		require.NoError(t, os.Symlink(linkSrcFileName, linkFileName))
> 
> 		_, err = WatchFile(ctx, linkFileName, c)
> 		require.NoError(t, err)
> 
> 		dirTwo := filepath.Join(dir, "dirTwo")
> 		require.NoError(t, os.Mkdir(dirTwo, os.ModePerm))
> 		fileOneBis := filepath.Join(dirTwo, "fileOne")
> 		f1bis, err := os.Create(fileOneBis)
> 		require.NoError(t, err)
> 		_, err = fmt.Fprintf(f1bis, "New file")
> 		require.NoError(t, err)
> 		require.NoError(t, f1bis.Close())
> 
> 		require.NoError(t, os.Remove(linkDirName))
> 		require.NoError(t, os.Symlink(dirTwo, linkDirName))
> 		assertRemove(t, <-c, linkFileName)
> 	})
> ```
> 
> Basically un kubernetes the symlink of file.txt points to .data/file.txt and .data is a symlink to .data-2021-11-07-... . When a change occurs K8S only changes the symlink .data if I am not wrong. Have you tested this with K8S ?

Face Palm... RTFM. Please ignore my last comment ;-)

StarAurryon:
@aeneasr @alnr waiting for feedback on the new implementation.

httpx/cert_reload.go
alnr: > I guess there's another one question buried here: How do we deal with the config value changing? So let's say `serve.tls.cert=file://abcd.pem` changes to `serve.tls.cert=file://not-abcd.pem`. How would we reload that with the current approach?


We should consider separating the notions of config hot reloading and TLS certificate reloading from an implementational POV. It's simpler to reason about a package/function which
1.  reads certs from a specific filesystem path
2.  reloads the files when they change
3.  allows the background reload to be cancelled

than it is to reason about a package which additionally also
4.  reads certificates from a base64 string (but only sometimes)
5.  generates a self-signed certificate (but only sometimes)
6.  changes its behavior completely during its lifetime

My proposal is to keep the interface of the cert reloader minimal. When the underlying config in Hydra/Kratos changes ("`serve.tls.cert=file://abcd.pem` changes to `serve.tls.cert=file://not-abcd.pem`"), the code in Hydra/Kratos should detect that, cancel the old TLS cert reloader, and spawn a new one pointing to the new path.

With a little bit of glue code in `github.com/ory/hydra/config` it should still be possible to do hot-reloading without closing the listening socket.

alnr:
For reference: https://github.com/ory/x/pull/502#discussion_r891093148

tlsx/cert_provider_test.go
```diff
@@ -0,0 +1,144 @@
+package tlsx
+
+// func TestCertProvider(t *testing.T) {
```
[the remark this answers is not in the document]
StarAurryon: Only waiting for the overall ok to modify initial tests

httpx/cert_reload.go
StarAurryon: My current implementation expect that on config reload you call LoadCertificates to perform the changes.

@aeneasr Please confirm that "serve.tls.cert=file://not-abcd.pem" is already handled by config handlers in hydra / kratos and does not impact.

tlsx/cert_provider.go
```diff
@@ -0,0 +1,181 @@
...
+		for {
+			select {
+			case <-p.ctx.Done():
+				return
+			case e, ok := <-p.fsEvent:
+				if !ok {
+					return
+				}
+
+				if err, isErr := e.(*watcherx.ErrorEvent); isErr {
```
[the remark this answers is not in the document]
StarAurryon: Done

tlsx/cert_provider_test.go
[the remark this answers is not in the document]
StarAurryon: Tests ready

tlsx/cert_provider.go
```diff
@@ -0,0 +1,181 @@
...
+type (
+	Provider interface {
+		SetCertificatesGenerator(CertificateGenerator) Provider
+		GetCertificate(hello *tls.ClientHelloInfo) (*tls.Certificate, error)
+		LoadCertificates(certString, keyString string, certPath, keyPath string) error
+	}
+
+	provider struct {
+		ctx    context.Context
+		logger *logrusx.Logger
```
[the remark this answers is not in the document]
StarAurryon: Moved to event channel instead as better to test and debug

alnr:
> @alnr Is it fine for you too ?

I'll have to defer to @aeneasr for the final OK. Personally, I still consider my comments [here](https://github.com/ory/x/pull/502#discussion_r891093148) applicable.

StarAurryon:
So only theses points remains for your mentioned comment:

> 3. allows the background reload to be cancelled
> 5. generates a self-signed certificate (but only sometimes)

For point 3 I don't think it is necessary, but why not.
For point 5, this is handled externally as ory hydra can use it's on hsm generation for certs.

alnr:
TLS hot-reloading was merged here: https://github.com/ory/x/pull/507

PR #509, opened 2022-05-17
Migration: WithTestdata

hperl:

This PR adds `WithTestdata` to the migrator and was extracted from https://github.com/ory/keto/blob/feat/uuid-mapping/internal/persistence/sql/migrations/migratest/migration_test.go#L34.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/popx/migration_box.go
+++ b/popx/migration_box.go
@@ -3,11 +3,14 @@ package popx
 import (
 	"io"
 	"io/fs"
+	"regexp"
 	"sort"
 	"strings"
+	"testing"
 
 	"github.com/gobuffalo/pop/v6"
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/require"
 
 	"github.com/ory/x/logrusx"
 )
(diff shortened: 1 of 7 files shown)


popx/migration_box.go
```diff
@@ -53,6 +55,57 @@ func WithGoMigrations(migrations Migrations) func(*MigrationBox) *MigrationBox {
 	}
 }
 
+// WithTestdata
+func WithTestdata(testdata fs.FS) func(*MigrationBox) *MigrationBox {
+	testdataPattern := regexp.MustCompile(`^(\d+)_testdata.sql`)
+	return func(m *MigrationBox) *MigrationBox {
```
zepatrik, line 61: Inside of the option function you don't have a `t *testing.T` parameter, but you can add it to `WithTestdata`. The `fs.WalkDir` collecting all the testdata does not have to happen in the returned function, but you could instead collect the migrations already outside of that and then just append them in the returned function. This will allow you to do `require.NoError(t, err)`.

popx/migration_box_gomigration_test.go
```diff
@@ -24,7 +24,7 @@ func TestGoMigrations(t *testing.T) {
 			Direction: "up",
 			Type:      "go",
 			DBType:    "all",
-			Runner: func(migration popx.Migration, _ *pop.Connection, tx *pop.Tx) error {
+			Runner: func(_ popx.Migration, _ *pop.Connection, _ *pop.Tx) error {
```
zepatrik, line 27: Alternative:
```suggestion
			Runner: func(popx.Migration, *pop.Connection, *pop.Tx) error {
```

popx/migration_box.go
```diff
@@ -92,7 +145,7 @@ func NewMigrationBox(dir fs.FS, m *Migrator, opts ...func(*MigrationBox) *Migrat
 		mb.Migrations[migration.Direction] = append(mb.Migrations[migration.Direction], migration)
 	}
 
-	return mb, nil
+	return mb, mb.check()
```
zepatrik, line 148: Both should work as callers always should check errors, but in case of an error it is cleaner to always return a zero value.
```suggestion
	if err := mb.check(); err != nil {
		return nil, err
	}
	return mb, nil
```

popx/migration_box.go
```diff
@@ -146,3 +199,24 @@ func (fm *MigrationBox) findMigrations(runner func([]byte) func(mf Migration, c
...
+		}
+	}
+	return false
+}
+
+// check checks that every "up" migration has a corresponding "down" migration.
+func (fm *MigrationBox) check() error {
+	for _, up := range fm.Migrations["up"] {
+		if !fm.hasDownMigrationWithVersion(up.Version) {
+			return fmt.Errorf("migration %s has no corresponding down migration", up.Version)
```
zepatrik, line 218:
```suggestion
			return errors.Errorf("migration %s has no corresponding down migration", up.Version)
```
Assuming the `github.com/pkg/errors` package. This error will automatically have a stack trace, which we inside of Ory always want to have.

popx/migration_box_testdata_test.go
```diff
@@ -0,0 +1,36 @@
...
+	require.NoError(t, c.Open())
+
+	mb, err := popx.NewMigrationBox(
+		empty,
+		popx.NewMigrator(c, logrusx.New("", ""), nil, 0),
+		popx.WithTestdata(testData))
+
+	require.NoError(t, err)
+	assert.Len(t, mb.Migrations["up"], 1)
+	assert.Equal(t, "testdata", mb.Migrations["up"][0].Name)
```
zepatrik, line 35: You could have some simple `CREATE TABLE` example in the migrations, with the testdata inserting a row. That would be a more realistic test. You can in the end verify whether the row was inserted.

zepatrik (changes requested):
Nice, just some minor comments :+1:

popx/migration_box.go
hperl: Ah yes, we talked about that :)

zepatrik (approved):

popx/migration_box.go
```diff
@@ -146,3 +203,24 @@ func (fm *MigrationBox) findMigrations(runner func([]byte) func(mf Migration, c
 		return nil
 	})
 }
+
+// hasDownMigrationWithVersion checks if there is a migration with the given
+// version.
+func (fm *MigrationBox) hasDownMigrationWithVersion(version string) bool {
```
grantzvolsky, line 209: This might be a bit too strict because the following migration set is valid: {`a.mysql.up.sql`, `a.postgres.up.sql`, `a.down.sql`}. This code that I wrote in Hydra may or may not be helpful: https://github.com/ory/hydra/blob/d6080061e91844fa4ba67eea1110ec84cb20ab4b/cmd/cli/handler_migrate.go#L200

grantzvolsky (commented):
@hperl Hi, the new validation may be a bit too strict, please see my comment

PR #510, opened 2022-05-17
fix: better sanitization of migration_table_name

hperl:

This PR ensures that the `migration_table_name` only contains alphanumerical characters and numbers and thus is safe to use in SQL string concatenation. Depending on the setup, the table name can be supplied as an URL parameter (e.g., `sqlite://file::memory:?_fk=true&migration_table_name=injection--`) and thus should better be sanitized before usage in SQL queries.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/popx/migrator.go
+++ b/popx/migrator.go
@@ -92,7 +92,7 @@ func (m *Migrator) UpTo(ctx context.Context, step int) (applied int, err error) 
 
 	c := m.Connection.WithContext(ctx)
 	err = m.exec(ctx, func() error {
-		mtn := m.migrationTableName(ctx, c)
+		mtn := m.sanitizedMigrationTableName(c)
 		mfs := m.Migrations["up"].SortAndFilter(c.Dialect.Name())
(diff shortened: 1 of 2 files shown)


popx/migrator.go
```diff
@@ -431,8 +431,8 @@ func (m MigrationStatuses) HasPending() bool {
 	return false
 }
 
-func (m *Migrator) migrationTableName(ctx context.Context, con *pop.Connection) string {
-	return con.MigrationTableName()
+func (m *Migrator) sanitizedMigrationTableName(con *pop.Connection) string {
+	return regexp.MustCompile(`\W`).ReplaceAllString(con.MigrationTableName(), "")
```
zepatrik, line 435: We could for performance reasons make the regexp a global, but I guess we don't use it during "normal" operation, but only when applying/checking migrations so it is not really of great concern.

zepatrik (approved):

PR #514, opened 2022-05-18
refactor: API changes for healthx for middleware

barelyhuman:
**BREAKING CHANGES**: changes the API definitions of the following 
`healthx/Alive, healthx/Ready, healthx/Version, healthx/SetHealthRoutes,healthx/SetVersionRoutes`

will add support for passing a middleware to the register
function, also changes the actual handlers to return http.handlers
instead

lays the base set of changes for https://github.com/ory/hydra/pull/3114

## Checklist

      and signed the CLA.
- [x] I have referenced an issue containing the design document if my change
      introduces a new feature.
      the changes.
      works.(modified existing tests)
      appropriate).

### DIFF
--- a/healthx/handler.go
+++ b/healthx/handler.go
@@ -79,17 +79,18 @@ func NewHandler(
 
 type router interface {
 	GET(path string, handle httprouter.Handle)
+	Handler(method, path string, handler http.Handler)
 }
 
 // SetHealthRoutes registers this handler's routes for health checking.
-func (h *Handler) SetHealthRoutes(r router, shareErrors bool) {
-	r.GET(AliveCheckPath, h.Alive)
-	r.GET(ReadyCheckPath, h.Ready(shareErrors))
+func (h *Handler) SetHealthRoutes(r router, shareErrors bool, middleware func(http.Handler) http.Handler) {
+	r.Handler("GET", AliveCheckPath, middleware(h.Alive()))
+	r.Handler("GET", ReadyCheckPath, middleware(h.Ready(shareErrors)))
 }
 
-// SetHealthRoutes registers this handler's routes for health checking.
-func (h *Handler) SetVersionRoutes(r router) {
-	r.GET(VersionPath, h.Version)
+// SetVersionRoutes registers this handler's routes for health checking.
+func (h *Handler) SetVersionRoutes(r router, middleware func(http.Handler) http.Handler) {
+	r.Handler("GET", VersionPath, middleware(h.Version()))
 }
 
 // Alive returns an ok status if the instance is ready to handle HTTP requests.
@@ -113,9 +114,11 @@ func (h *Handler) SetVersionRoutes(r router) {
 //     Responses:
 //       200: healthStatus
 //       500: genericError
-func (h *Handler) Alive(rw http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-	h.H.Write(rw, r, &swaggerHealthStatus{
-		Status: "ok",
+func (h *Handler) Alive() http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		h.H.Write(rw, r, &swaggerHealthStatus{
+			Status: "ok",
+		})
 	})
 }
 
@@ -140,8 +143,8 @@ func (h *Handler) Alive(rw http.ResponseWriter, r *http.Request, _ httprouter.Pa
 //     Responses:
 //       200: healthStatus
 //       503: healthNotReadyStatus
-func (h *Handler) Ready(shareErrors bool) httprouter.Handle {
-	return func(rw http.ResponseWriter, r *http.Request, _ httprouter.Params) {
+func (h *Handler) Ready(shareErrors bool) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
 		var notReady = swaggerNotReadyStatus{
 			Errors: map[string]string{},
 		}
@@ -164,7 +167,7 @@ func (h *Handler) Ready(shareErrors bool) httprouter.Handle {
 		h.H.Write(rw, r, &swaggerHealthStatus{
 			Status: "ok",
 		})
-	}
+	})
 }
 
 // Version returns this service's versions.
@@ -186,8 +189,10 @@ func (h *Handler) Ready(shareErrors bool) httprouter.Handle {
 //
 //	   Responses:
 // 			200: version
-func (h *Handler) Version(rw http.ResponseWriter, r *http.Request, _ httprouter.Params) {
-	h.H.Write(rw, r, &swaggerVersion{
-		Version: h.VersionString,
+func (h *Handler) Version() http.HandlerFunc {
+	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		h.H.Write(rw, r, &swaggerVersion{
+			Version: h.VersionString,
+		})
 	})
 }
--- a/healthx/handler.go
+++ b/healthx/handler.go
@@ -62,9 +62,11 @@ type Handler struct {
 	ReadyChecks   ReadyCheckers
 }
 
-type Options struct {
-	Middleware func(http.Handler) http.Handler
+type options struct {
+	middleware func(http.Handler) http.Handler
 }
+
+type Options func(*options)
 
 // NewHandler instantiates a handler.
 func NewHandler(
@@ -84,26 +86,26 @@ type router interface {
 }
 
 // SetHealthRoutes registers this handler's routes for health checking.
-func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...func(*Options)) {
-	o := &Options{}
+func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Options) {
+	o := &options{}
 
 	for _, opt := range opts {
 		opt(o)
 	}
 
-	r.Handler("GET", AliveCheckPath, o.Middleware(h.Alive()))
-	r.Handler("GET", ReadyCheckPath, o.Middleware(h.Ready(shareErrors)))
+	r.Handler("GET", AliveCheckPath, o.middleware(h.Alive()))
+	r.Handler("GET", ReadyCheckPath, o.middleware(h.Ready(shareErrors)))
 }
 
 // SetVersionRoutes registers this handler's routes for health checking.
-func (h *Handler) SetVersionRoutes(r router, opts ...func(*Options)) {
-	o := &Options{}
+func (h *Handler) SetVersionRoutes(r router, opts ...Options) {
+	o := &options{}
 
 	for _, opt := range opts {
 		opt(o)
 	}
 
-	r.Handler("GET", VersionPath, o.Middleware(h.Version()))
+	r.Handler("GET", VersionPath, o.middleware(h.Version()))
 }
 
 // Alive returns an ok status if the instance is ready to handle HTTP requests.
@@ -208,4 +210,12 @@ func (h *Handler) Version() http.Handler {
 			Version: h.VersionString,
 		})
 	})
+}
+
+// WithMiddleware accepts a http.Handler to be run on the
+// route handlers
+func WithMiddleware(h func(http.Handler) http.Handler) Options {
+	return func(o *options) {
+		o.middleware = h
+	}
 }
--- a/healthx/handler.go
+++ b/healthx/handler.go
@@ -62,6 +62,10 @@ type Handler struct {
 	ReadyChecks   ReadyCheckers
 }
 
+type Options struct {
+	Middleware func(http.Handler) http.Handler
+}
+
 // NewHandler instantiates a handler.
 func NewHandler(
 	h herodot.Writer,
@@ -79,29 +83,27 @@ type router interface {
 	Handler(method, path string, handler http.Handler)
 }
 
-type Middleware func(http.Handler) http.Handler
-
 // SetHealthRoutes registers this handler's routes for health checking.
-func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Middleware) {
-	aliveHandler := h.Alive()
-	readyHandler := h.Ready(shareErrors)
+func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...func(*Options)) {
+	o := &Options{}
 
 	for _, opt := range opts {
-		aliveHandler = opt(aliveHandler)
-		readyHandler = opt(readyHandler)
+		opt(o)
 	}
 
-	r.Handler("GET", AliveCheckPath, aliveHandler)
-	r.Handler("GET", ReadyCheckPath, readyHandler)
+	r.Handler("GET", AliveCheckPath, o.Middleware(h.Alive()))
+	r.Handler("GET", ReadyCheckPath, o.Middleware(h.Ready(shareErrors)))
 }
 
 // SetVersionRoutes registers this handler's routes for health checking.
-func (h *Handler) SetVersionRoutes(r router, opts ...Middleware) {
-	versionHandler := h.Version()
+func (h *Handler) SetVersionRoutes(r router, opts ...func(*Options)) {
+	o := &Options{}
+
 	for _, opt := range opts {
-		versionHandler = opt(versionHandler)
+		opt(o)
 	}
-	r.Handler("GET", VersionPath, versionHandler)
+
+	r.Handler("GET", VersionPath, o.Middleware(h.Version()))
 }
 
 // Alive returns an ok status if the instance is ready to handle HTTP requests.
--- a/healthx/handler.go
+++ b/healthx/handler.go
@@ -88,24 +88,36 @@ type router interface {
 // SetHealthRoutes registers this handler's routes for health checking.
 func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Options) {
 	o := &options{}
+	aliveHandler := h.Alive()
+	readyHandler := h.Ready(shareErrors)
 
 	for _, opt := range opts {
 		opt(o)
 	}
 
-	r.Handler("GET", AliveCheckPath, o.middleware(h.Alive()))
-	r.Handler("GET", ReadyCheckPath, o.middleware(h.Ready(shareErrors)))
+	if o.middleware != nil {
+		aliveHandler = o.middleware(aliveHandler)
+		readyHandler = o.middleware(readyHandler)
+	}
+
+	r.Handler("GET", AliveCheckPath, aliveHandler)
+	r.Handler("GET", ReadyCheckPath, readyHandler)
 }
 
 // SetVersionRoutes registers this handler's routes for health checking.
 func (h *Handler) SetVersionRoutes(r router, opts ...Options) {
 	o := &options{}
+	versionHandler := h.Version()
 
 	for _, opt := range opts {
 		opt(o)
 	}
 
-	r.Handler("GET", VersionPath, o.middleware(h.Version()))
+	if o.middleware != nil {
+		versionHandler = o.middleware(versionHandler)
+	}
+
+	r.Handler("GET", VersionPath, versionHandler)
 }
 
 // Alive returns an ok status if the instance is ready to handle HTTP requests.
--- a/healthx/handler.go
+++ b/healthx/handler.go
@@ -23,8 +23,6 @@ package healthx
 import (
 	"net/http"
 
-	"github.com/julienschmidt/httprouter"
-
 	"github.com/ory/herodot"
 )
 
@@ -78,19 +76,32 @@ func NewHandler(
 }
 
 type router interface {
-	GET(path string, handle httprouter.Handle)
 	Handler(method, path string, handler http.Handler)
 }
 
+type Middleware func(http.Handler) http.Handler
+
 // SetHealthRoutes registers this handler's routes for health checking.
-func (h *Handler) SetHealthRoutes(r router, shareErrors bool, middleware func(http.Handler) http.Handler) {
-	r.Handler("GET", AliveCheckPath, middleware(h.Alive()))
-	r.Handler("GET", ReadyCheckPath, middleware(h.Ready(shareErrors)))
+func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Middleware) {
+	aliveHandler := h.Alive()
+	readyHandler := h.Ready(shareErrors)
+
+	for _, opt := range opts {
+		aliveHandler = opt(aliveHandler)
+		readyHandler = opt(readyHandler)
+	}
+
+	r.Handler("GET", AliveCheckPath, aliveHandler)
+	r.Handler("GET", ReadyCheckPath, readyHandler)
 }
 
 // SetVersionRoutes registers this handler's routes for health checking.
-func (h *Handler) SetVersionRoutes(r router, middleware func(http.Handler) http.Handler) {
-	r.Handler("GET", VersionPath, middleware(h.Version()))
+func (h *Handler) SetVersionRoutes(r router, opts ...Middleware) {
+	versionHandler := h.Version()
+	for _, opt := range opts {
+		versionHandler = opt(versionHandler)
+	}
+	r.Handler("GET", VersionPath, versionHandler)
 }
 
 // Alive returns an ok status if the instance is ready to handle HTTP requests.
@@ -189,7 +200,7 @@ func (h *Handler) Ready(shareErrors bool) http.Handler {
 //
 //	   Responses:
 // 			200: version
-func (h *Handler) Version() http.HandlerFunc {
+func (h *Handler) Version() http.Handler {
 	return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
 		h.H.Write(rw, r, &swaggerVersion{
 			Version: h.VersionString,
--- a/healthx/handler_test.go
+++ b/healthx/handler_test.go
@@ -39,21 +39,9 @@ import (
 const mockHeaderKey = "middleware-header"
 const mockHeaderValue = "test-header-value"
 
-// MiddlewareHeader is a example logger for the below tests
-type MiddlewareHeader struct {
-	handler http.Handler
-	t       *testing.T
-}
-
-func (l *MiddlewareHeader) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	w.Header().Add(mockHeaderKey, mockHeaderValue)
-	l.handler.ServeHTTP(w, r)
-}
-
-func loggerMiddleware(t *testing.T) func(n http.Handler) http.Handler {
-	return func(n http.Handler) http.Handler {
-		logger := &MiddlewareHeader{handler: n, t: t}
-		return logger
+func WithMiddleware(h func(http.Handler) http.Handler) func(*Options) {
+	return func(o *Options) {
+		o.Middleware = h
 	}
 }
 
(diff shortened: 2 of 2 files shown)


healthx/handler.go
```diff
@@ -79,17 +79,18 @@ func NewHandler(
...
+	Handler(method, path string, handler http.Handler)
 }
 
 // SetHealthRoutes registers this handler's routes for health checking.
-func (h *Handler) SetHealthRoutes(r router, shareErrors bool) {
-	r.GET(AliveCheckPath, h.Alive)
-	r.GET(ReadyCheckPath, h.Ready(shareErrors))
+func (h *Handler) SetHealthRoutes(r router, shareErrors bool, middleware func(http.Handler) http.Handler) {
+	r.Handler("GET", AliveCheckPath, middleware(h.Alive()))
+	r.Handler("GET", ReadyCheckPath, middleware(h.Ready(shareErrors)))
```
zepatrik, line 88: In most cases we don't want to set the middleware. I suggest we use the options pattern here instead.
```suggestion
func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Options) {
	o := getOptions(opts)
	r.Handler("GET", AliveCheckPath, o.middleware(h.Alive()))
	r.Handler("GET", ReadyCheckPath, o.middleware(h.Ready(shareErrors)))
```
Per default the middleware would be a noop then.

zepatrik: By that we also avoid introducing a breaking change.

healthx/handler.go
```diff
@@ -79,17 +79,18 @@ func NewHandler(
 
 type router interface {
 	GET(path string, handle httprouter.Handle)
+	Handler(method, path string, handler http.Handler)
```
zepatrik, line 82:
```suggestion
	Handler(method, path string, handler http.Handler)
```

healthx/handler_test.go
```diff
@@ -47,9 +65,10 @@ func TestHealth(t *testing.T) {
 			},
 		},
 	}
+
 	router := httprouter.New()
-	handler.SetHealthRoutes(router, true)
-	handler.SetVersionRoutes(router)
+	handler.SetHealthRoutes(router, true, loggerMiddleware)
+	handler.SetVersionRoutes(router, loggerMiddleware)
```
zepatrik, line 71: It would be good to add some assertions here instead of logging. Logs are not shown in tests without `-v`, and not checked by automation.

zepatrik (changes requested):

healthx/handler.go
barelyhuman: Thanks for the tip, will update this

healthx/handler_test.go
barelyhuman: these were just to satisfy the tests for the above change, in case the above changes wouldn't get approved.
Will add needed assertions with in the next commit

healthx/handler.go
barelyhuman: Though, I was thinking, more like this.

if it makes sense

```go
type Middleware func(http.Handler) http.Handler

// SetHealthRoutes registers this handler's routes for health checking.
func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Middleware) {
	aliveHandler := h.Alive()
	readyHandler := h.Ready(shareErrors)

	for _, opt := range opts {
		aliveHandler = opt(aliveHandler)
	}

	for _, opt := range opts {
		readyHandler = opt(readyHandler)
	}

	r.Handler("GET", AliveCheckPath, aliveHandler)
	r.Handler("GET", ReadyCheckPath, readyHandler)
}
```

healthx/handler_test.go
```diff
@@ -36,6 +36,27 @@ import (
...
+	w.Header().Add(mockHeaderKey, mockHeaderValue)
+	l.handler.ServeHTTP(w, r)
+}
+
+func loggerMiddleware(t *testing.T) func(n http.Handler) http.Handler {
+	return func(n http.Handler) http.Handler {
+		logger := &MiddlewareHeader{handler: n, t: t}
+		return logger
+	}
+}
```
Benehiko, line 58: I don't think you need the `MiddlewareHeader` struct here, we could just wrap it in the middleware immediately functionally and then just do an assertion as well to make sure the request is coming from the right source, e.g. the method is a GET, or pass the expected values. You can also still set the mock header values to assert later that it did indeed execute.

```go
writer.Header().Add(mockHeaderKey, mockHeaderValue)
```

I would just do 

```go
testMiddleware := func(t *testing.T, expectedStuff func(*testing.T, http.ResponseWriter,*http.Request)) func(next http.Handler) http.Handler { 
  //call expectedStuff
  //do middleware stuff
  })
}
```

I would also call this something else instead of `loggerMiddleware`. Maybe `testMiddleware` since it is a test middleware for our health and version routes.

Check out https://github.com/ory/x/blob/4bf4fdd378ce28464cc6374314b9026b2291a486/proxy/proxy.go#L201-L217 for inspiration on how to do this.

Benehiko (changes requested):

healthx/handler.go
```diff
@@ -78,18 +76,32 @@ func NewHandler(
...
 }
 
+type Middleware func(http.Handler) http.Handler
+
 // SetHealthRoutes registers this handler's routes for health checking.
-func (h *Handler) SetHealthRoutes(r router, shareErrors bool) {
-	r.GET(AliveCheckPath, h.Alive)
-	r.GET(ReadyCheckPath, h.Ready(shareErrors))
+func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Middleware) {
```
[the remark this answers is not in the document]
barelyhuman: or are we talking about creating something similar to the proxy options setup you mentioned below?

healthx/handler_test.go
barelyhuman: > I would just do

Um, while it works , in my opinion the one with the type struct is a little more readable?
the function as a param and return do add a lot of noise to the simple definition of a http.handler

barelyhuman: Here's what the working code would look like with the above change 
```go
```

healthx/handler.go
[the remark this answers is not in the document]
Benehiko: yes exactly, the options shouldn't just be a `http.Handlers`. Use the proxy as an example.

healthx/handler_test.go
Benehiko: you only need one function :) not 2

Benehiko: > Um, while it works , in my opinion the one with the type struct is a little more readable?

I don't agree, with the struct you have way more lines of code and you still need the middleware function anyway. with this change you don't need the struct and can do it in one function that does the assertions (which is what we want to test for in the middleware) and a test asserting the middleware was used.

healthx/handler.go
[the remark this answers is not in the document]
barelyhuman: Yeah, I have the base idea of implementing it, 
I wanted to confirm a requirement, do we wanna do multiple middlewares like  if someone was to write, he'd write it 
```go
WithMiddleware(
func1,
func2 
)
```

or do we wanna do single middleware and the developer decides if he wants to wrap them up

```go 
withMiddleware(func1)
```

based on that the code will change as below (you obviously know, but since it's also an assignment, just confirming) 

```go
type Options struct {
	Middleware func(http.Handler) http.Handler
}

// SetHealthRoutes registers this handler's routes for health checking.
func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...func(*Options)) {
	o := &Options{}

	for _, opt := range opts {
		opt(o)
	}

	r.Handler("GET", AliveCheckPath, o.Middleware(h.Alive()))
	r.Handler("GET", ReadyCheckPath, o.Middleware(h.Ready(shareErrors)))
}
```

vs 

```go
func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...func(*Options)) {
	o := &Options{}

	for _, opt := range opts {
		opt(o)
	}

	aliveHandler := h.Alive()
	readyHandler := h.Ready(shareErrors)

	for _, middleware := range o.Middleware {
		aliveHandler = middleware(aliveHandler)
		readyHandler = middleware(readyHandler)
	}

	r.Handler("GET", AliveCheckPath, aliveHandler)
	r.Handler("GET", ReadyCheckPath, readyHandler)
}
```

I prefer the second one cause then I can do the following 
```go
handler.SetHealthRoutes(router, true,
		WithMiddleware(
                   addMockHeaderMiddleware, 
                   testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
			assert.Equal(t, "GET", r.Method)
		})),
	)
```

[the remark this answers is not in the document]
barelyhuman: @Benehiko Will push the needed code as soon as you confirm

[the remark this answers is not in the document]
Benehiko: no i think you can keep it a single middleware since you can always wrap more middleware over the one you are passing down to the healthx package. The order of execution for middleware is sometimes really important, so I think it makes more sense to not implement something like that which could lead to future bugs because one middleware is executing after another instead of before.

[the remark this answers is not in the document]
barelyhuman: 👍🏼 understood

healthx/handler.go
```diff
@@ -64,6 +62,10 @@ type Handler struct {
 	ReadyChecks   ReadyCheckers
 }
 
+type Options struct {
+	Middleware func(http.Handler) http.Handler
+}
```
Benehiko, line 67: Why is this a struct instead of a function?

```go
type Options func(*options)
```

Wouldn't it then be easier to add options through functions? With this method we expose the `Middleware` struct field to the outside - which is probably not what we want?

e.g.

```go
WIthMiddleware(m http.Handler) Options {
  return func(o *options) {
    o.middleware = m
  }
}

File: healthx/handler_test.go [hunk 7]
```diff
@@ -36,6 +36,15 @@ import (
...
 )
 
+const mockHeaderKey = "middleware-header"
+const mockHeaderValue = "test-header-value"
+
+func WithMiddleware(h func(http.Handler) http.Handler) func(*Options) {
+	return func(o *Options) {
+		o.Middleware = h
+	}
+}
```
Benehiko, line 46: why in the `_test` file instead of inside the `handler`?

Benehiko (changes requested):

healthx/handler_test.go
barelyhuman: Only needed it in the test
might act as an example for anyone who would want to implement something similar 
also, so that I wouldn't have to combine it all in one function for the test
 at least, that was my chain of thought

healthx/handler.go
barelyhuman: There isn't a `options` struct in the healthx, I'd still have to write a struct for `options` , just did it on the original struct.

As for exposing the type, need a confirmation.
`Options` being exposed is fine (cause well you need the reference in the used functions), and `Middleware` isn't, right? Will be changed. 

Thought, I might need an explanation as to why exposing it would be a bad decision, if you don't mind that is

healthx/handler.go
```diff
@@ -78,18 +82,30 @@ func NewHandler(
...
-	r.GET(VersionPath, h.Version)
+// SetVersionRoutes registers this handler's routes for health checking.
+func (h *Handler) SetVersionRoutes(r router, opts ...Options) {
+	o := &options{}
+
+	for _, opt := range opts {
+		opt(o)
+	}
+
+	r.Handler("GET", VersionPath, o.middleware(h.Version()))
```
Benehiko, line 108: so what happens when `o.middleware` is nil?

healthx/handler.go
```diff
@@ -78,18 +82,30 @@ func NewHandler(
...
-	r.GET(ReadyCheckPath, h.Ready(shareErrors))
+func (h *Handler) SetHealthRoutes(r router, shareErrors bool, opts ...Options) {
+	o := &options{}
+
+	for _, opt := range opts {
+		opt(o)
+	}
+
+	r.Handler("GET", AliveCheckPath, o.middleware(h.Alive()))
+	r.Handler("GET", ReadyCheckPath, o.middleware(h.Ready(shareErrors)))
```
Benehiko, line 97: so what happens when `o.middleware` is nil?

healthx/handler_test.go
```diff
@@ -47,9 +50,30 @@ func TestHealth(t *testing.T) {
...
+
+	handler.SetHealthRoutes(router, true,
+		WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
+			assert.Equal(t, "GET", r.Method)
+		})),
+	)
+
+	handler.SetVersionRoutes(router, WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "GET", r.Method)
+	})))
```
Benehiko, line 75: Could we maybe also add a test to make sure these handlers will still work without a middleware?

Benehiko: We can separate them using a sub test 

```go
t.Run("case=with middleware", func(t *testing.T){
  ...test
})
```

barelyhuman: want all the tests to be inside the `t.Run` ?
It'd be better with just them being in 2 different TestHealth functions?
one for with and one for without, instead of both being the same TestHealth function

barelyhuman: Here's a patch before I commit, doesn't make sense to keep adding commits if it's not up to mark

```diff
diff --git a/healthx/handler_test.go b/healthx/handler_test.go
index 2d4fe82..461b1a0 100644
--- a/healthx/handler_test.go
+++ b/healthx/handler_test.go
@@ -40,73 +40,125 @@ const mockHeaderKey = "middleware-header"
 const mockHeaderValue = "test-header-value"
 
 func TestHealth(t *testing.T) {
-	alive := errors.New("not alive")
-	handler := &Handler{
-		H:             herodot.NewJSONWriter(nil),
-		VersionString: "test version",
-		ReadyChecks: map[string]ReadyChecker{
-			"test": func(r *http.Request) error {
-				return alive
+	t.Run("case:without middleware", func(t *testing.T) {
+		// separate set of `handler` and `alive` as the `alive` would need to be reset after the last test in the above
+		alive := errors.New("not alive")
+		handler := &Handler{
+			H:             herodot.NewJSONWriter(nil),
+			VersionString: "test version",
+			ReadyChecks: map[string]ReadyChecker{
+				"test": func(r *http.Request) error {
+					return alive
+				},
 			},
-		},
-	}
-
-	// middleware to assert and test before the request is completed
-	testMiddleware := func(t *testing.T, assertFunc func(*testing.T, http.ResponseWriter, *http.Request)) func(next http.Handler) http.Handler {
-		return func(h http.Handler) http.Handler {
-			return http.HandlerFunc(func(writer http.ResponseWriter, req *http.Request) {
-				writer.Header().Add(mockHeaderKey, mockHeaderValue)
-				assertFunc(t, writer, req)
-				h.ServeHTTP(writer, req)
-			})
 		}
-	}
 
-	router := httprouter.New()
+		router := httprouter.New()
 
-	handler.SetHealthRoutes(router, true,
-		WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
+		handler.SetHealthRoutes(router, true)
+		handler.SetVersionRoutes(router)
+
+		ts := httptest.NewServer(router)
+
+		c := http.DefaultClient
+
+		var healthBody swaggerHealthStatus
+		response, err := c.Get(ts.URL + AliveCheckPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusOK, response.StatusCode)
+		require.NoError(t, json.NewDecoder(response.Body).Decode(&healthBody))
+		assert.EqualValues(t, "ok", healthBody.Status)
+
+		var versionBody swaggerVersion
+		response, err = c.Get(ts.URL + VersionPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusOK, response.StatusCode)
+		require.NoError(t, json.NewDecoder(response.Body).Decode(&versionBody))
+		require.EqualValues(t, versionBody.Version, handler.VersionString)
+
+		response, err = c.Get(ts.URL + ReadyCheckPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusServiceUnavailable, response.StatusCode)
+		out, err := ioutil.ReadAll(response.Body)
+		require.NoError(t, err)
+		assert.EqualValues(t, "ok", healthBody.Status)
+		assert.Equal(t, `{"errors":{"test":"not alive"}}`, strings.TrimSpace(string(out)))
+
+		alive = nil
+		response, err = c.Get(ts.URL + ReadyCheckPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusOK, response.StatusCode)
+		require.NoError(t, json.NewDecoder(response.Body).Decode(&versionBody))
+		require.EqualValues(t, versionBody.Version, handler.VersionString)
+
+	})
+
+	t.Run("case:with middleware", func(t *testing.T) {
+		// separate set of `handler` and `alive` as the `alive` would need to be reset after the last test in the above
+		alive := errors.New("not alive")
+		handler := &Handler{
+			H:             herodot.NewJSONWriter(nil),
+			VersionString: "test version",
+			ReadyChecks: map[string]ReadyChecker{
+				"test": func(r *http.Request) error {
+					return alive
+				},
+			},
+		}
+
+		// middleware to assert and test before the request is completed
+		testMiddleware := func(t *testing.T, assertFunc func(*testing.T, http.ResponseWriter, *http.Request)) func(next http.Handler) http.Handler {
+			return func(h http.Handler) http.Handler {
+				return http.HandlerFunc(func(writer http.ResponseWriter, req *http.Request) {
+					writer.Header().Add(mockHeaderKey, mockHeaderValue)
+					assertFunc(t, writer, req)
+					h.ServeHTTP(writer, req)
+				})
+			}
+		}
+
+		routerWithMiddleware := httprouter.New()
+
+		ts := httptest.NewServer(routerWithMiddleware)
+		c := http.DefaultClient
+
+		handler.SetHealthRoutes(routerWithMiddleware, true,
+			WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
+				assert.Equal(t, "GET", r.Method)
+			})),
+		)
+
+		handler.SetVersionRoutes(routerWithMiddleware, WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
 			assert.Equal(t, "GET", r.Method)
-		})),
-	)
-
-	handler.SetVersionRoutes(router, WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
-		assert.Equal(t, "GET", r.Method)
-	})))
-
-	ts := httptest.NewServer(router)
-	c := http.DefaultClient
-
-	var healthBody swaggerHealthStatus
-	response, err := c.Get(ts.URL + AliveCheckPath)
-	require.NoError(t, err)
-	require.EqualValues(t, http.StatusOK, response.StatusCode)
-	require.NoError(t, json.NewDecoder(response.Body).Decode(&healthBody))
-	assert.EqualValues(t, "ok", healthBody.Status)
-	assert.EqualValues(t, mockHeaderValue, response.Header.Get(mockHeaderKey))
-
-	var versionBody swaggerVersion
-	response, err = c.Get(ts.URL + VersionPath)
-	require.NoError(t, err)
-	require.EqualValues(t, http.StatusOK, response.StatusCode)
-	require.NoError(t, json.NewDecoder(response.Body).Decode(&versionBody))
-	require.EqualValues(t, versionBody.Version, handler.VersionString)
-	assert.EqualValues(t, mockHeaderValue, response.Header.Get(mockHeaderKey))
-
-	response, err = c.Get(ts.URL + ReadyCheckPath)
-	require.NoError(t, err)
-	require.EqualValues(t, http.StatusServiceUnavailable, response.StatusCode)
-	out, err := ioutil.ReadAll(response.Body)
-	require.NoError(t, err)
-	assert.EqualValues(t, "ok", healthBody.Status)
-	assert.Equal(t, `{"errors":{"test":"not alive"}}`, strings.TrimSpace(string(out)))
-	assert.EqualValues(t, mockHeaderValue, response.Header.Get(mockHeaderKey))
-
-	alive = nil
-	response, err = c.Get(ts.URL + ReadyCheckPath)
-	require.NoError(t, err)
-	require.EqualValues(t, http.StatusOK, response.StatusCode)
-	require.NoError(t, json.NewDecoder(response.Body).Decode(&versionBody))
-	require.EqualValues(t, versionBody.Version, handler.VersionString)
-	assert.EqualValues(t, mockHeaderValue, response.Header.Get(mockHeaderKey))
+		})))
+
+		var healthBody swaggerHealthStatus
+		response, err := c.Get(ts.URL + AliveCheckPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusOK, response.StatusCode)
+		require.NoError(t, json.NewDecoder(response.Body).Decode(&healthBody))
+		assert.EqualValues(t, "ok", healthBody.Status)
+
+		var versionBody swaggerVersion
+		response, err = c.Get(ts.URL + VersionPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusOK, response.StatusCode)
+		require.NoError(t, json.NewDecoder(response.Body).Decode(&versionBody))
+		require.EqualValues(t, versionBody.Version, handler.VersionString)
+
+		response, err = c.Get(ts.URL + ReadyCheckPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusServiceUnavailable, response.StatusCode)
+		out, err := ioutil.ReadAll(response.Body)
+		require.NoError(t, err)
+		assert.EqualValues(t, "ok", healthBody.Status)
+		assert.Equal(t, `{"errors":{"test":"not alive"}}`, strings.TrimSpace(string(out)))
+
+		alive = nil
+		response, err = c.Get(ts.URL + ReadyCheckPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusOK, response.StatusCode)
+		require.NoError(t, json.NewDecoder(response.Body).Decode(&versionBody))
+		require.EqualValues(t, versionBody.Version, handler.VersionString)
+	})
 }

```

healthx/handler_test.go
```diff
@@ -36,49 +36,133 @@ import (
...
+					h.ServeHTTP(writer, req)
+				})
+			}
+		}
+
+		router := httprouter.New()
+
+		handler.SetHealthRoutes(router, true,
+			WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
+				assert.Equal(t, "GET", r.Method)
```
Benehiko, line 124: Maybe we can add more assertions here, such as request path and request headers.

healthx/handler_test.go
```diff
@@ -36,49 +36,133 @@ import (
...
+			})),
+		)
+
+		handler.SetVersionRoutes(router, WithMiddleware(testMiddleware(t, func(t *testing.T, rw http.ResponseWriter, r *http.Request) {
+			assert.Equal(t, "GET", r.Method)
```
Benehiko, line 129: same here

healthx/handler_test.go
```diff
@@ -36,49 +36,133 @@ import (
...
+		require.NoError(t, err)
+		assert.EqualValues(t, "ok", healthBody.Status)
+		assert.Equal(t, `{"errors":{"test":"not alive"}}`, strings.TrimSpace(string(out)))
+
+		alive = nil
+		response, err = c.Get(ts.URL + ReadyCheckPath)
+		require.NoError(t, err)
+		require.EqualValues(t, http.StatusOK, response.StatusCode)
+		require.NoError(t, json.NewDecoder(response.Body).Decode(&versionBody))
+		require.EqualValues(t, versionBody.Version, handler.VersionString)
```
Benehiko, line 92: maybe we can abstract this into a helper function that is reused in the test below since they essentially should have the same assertions on basic functionality except for the middleware adding a header.
Let's keep the code DRY ;)

Benehiko (changes requested):
Already looks a lot better :+1: :)

healthx/handler_test.go
barelyhuman: done, let me know if you need something changed in that

healthx/handler_test.go
```diff
@@ -36,8 +36,12 @@ import (
 	"github.com/ory/herodot"
 )
 
-func TestHealth(t *testing.T) {
+const mockHeaderKey = "middleware-header"
+const mockHeaderValue = "test-header-value"
+
+func coreHealthTests(t *testing.T, withMiddleware bool) {
```
Benehiko, line 42: Just a bit of a nit
I just wanted these lines https://github.com/ory/x/pull/514/files#diff-519494e59fcbc1e8e99737385005a0c4b843639063e9ad54f4a22d3ba3baf646R98-R129 abstracted then you will always just call like so instead of doing some if statement logic which makes it more difficult to read in future.

```go
t.Run("case:without middleware", func(t *testing.T) {
    // do handler stuff
    assertDefaultResponse(t, false)
})

t.Run("case:with middleware", func(t *testing.T) {
    // do middleware stuff with health and version
    assertDefaultResponse(t)
    // assert middleware response
})
```

Benehiko (changes requested):
Sorry to ask more changes, I think it's really almost done. 
Just a small nit comment and then we are good to go :)

healthx/handler_test.go
barelyhuman: Done, 
Though, maybe the approach I've used is not what you were thinking. 

Also, a tiny doubt 
https://github.com/ory/x/pull/514/commits/5ffdf2020e90ed3de88338572c859e9615c7c7a7#diff-519494e59fcbc1e8e99737385005a0c4b843639063e9ad54f4a22d3ba3baf646R46-R48

I'm modifying the handler internally since I didn't wanna pass the `error` typed variable around, unspecific to the test, doesn't that mean I'm modifying behaviour inside the abstraction and that's not right?

Benehiko:
Hey sorry for only getting back to you now, Monday was a bit hectic :sweat_smile: 

So the PR in general looks good, however, the structure of the tests don't look too good to me. 
Especially on something like https://github.com/ory/x/pull/514/files#diff-519494e59fcbc1e8e99737385005a0c4b843639063e9ad54f4a22d3ba3baf646R54-R59

It's not good to have many if statements and things in the tests when you want to assert something. It is always good to be explicit on a failure or explicit on an expected result. In the end we still ended up with duplication here https://github.com/ory/x/pull/514/files#diff-519494e59fcbc1e8e99737385005a0c4b843639063e9ad54f4a22d3ba3baf646R65-R96. 

The purpose of the helper function is to help with asserting our responses across the test (e.g. middleware and non-middleware). These responses should stay the same in this case as the middleware isn't modifying any response.

I thought it would be easier to quickly add the changes here to the PR so we could get it merged. While adding the changes I also realised some of the tests were actually incorrectly written and the original tests were someitmes reusing the responses from previous http requests. The objects were also reused in the different expected responses and asserted values incorrectly. 

https://github.com/ory/x/blob/master/healthx/handler_test.go#L56-L83

barelyhuman:
I understand the requirement, though I've been a bit anxious about making any major changes since that might not be what you are looking for, if you want the tests to be more reusable, I'll do so but it might lead to more change comments and if that's okay, I'll go ahead and do so

healthx/handler_test.go
```diff
@@ -37,48 +37,172 @@ import (
...
+
+		ts := httptest.NewServer(router)
+		defer ts.Close()
+
+		handler.SetHealthRoutes(router, true)
+		handler.SetVersionRoutes(router)
+
+		for _, tc := range testCases {
+			t.Run("case="+tc.description, func(t *testing.T) {
+				tc.test(t, tc.url(ts.URL), handler)
```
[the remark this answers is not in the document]
Benehiko: make sense :+1:

vinckr:
Hello @barelyhuman 
Congrats on merging your first PR in Ory 🎉 !
Your contribution will soon be helping secure millions of identities around the globe 🌏. 
As a small token of appreciation we send all our first time contributors a gift package to welcome them to the community. 
[Please drop me an email and I will forward you the form to claim your Ory swag!](mailto:vincent@ory.sh)

PR #515, opened 2022-05-19
test: ensure that down migrations are considered

hperl:
This PR adds more tests to check that down migrations are properly found.

### DIFF
--- a/popx/migration_box_testdata_test.go
+++ b/popx/migration_box_testdata_test.go
@@ -19,6 +19,9 @@ var testData embed.FS
 //go:embed stub/migrations/testdata_migrations/*
 var empty embed.FS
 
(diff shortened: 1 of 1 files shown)


hperl:
@grantzvolsky as discussed :)

popx/migration_box_testdata_test.go
```diff
@@ -46,3 +49,20 @@ func TestMigrationBoxWithTestdata(t *testing.T) {
 	pop.Debug = false
 	assert.Equal(t, "testdata", data.Data)
 }
+
+func TestMigrationBox_CheckNoErr(t *testing.T) {
```
zepatrik, line 53: Was there no code change required?

hperl: No, it works as intended.

PR #536, opened 2022-06-14
fix: deduplicate changefeed events

hperl:
Cockroach might send change events multiple times, so we need
to deduplicate them.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/watcherx/changefeed.go
+++ b/watcherx/changefeed.go
@@ -2,6 +2,7 @@ package watcherx
 
 import (
 	"context"
+	"crypto/sha256"
 	"database/sql"
 	"fmt"
 	"strings"
@@ -68,7 +69,10 @@ const heartBeatInterval = time.Second
 // Watcher.DispatchNow() does not have an effect in this method.
 //
 // This watcher is blocking to allow proper context cancellation and clean up.
-func WatchChangeFeed(ctx context.Context, cx *sqlx.DB, tableName string, c EventChannel, cursor time.Time) (_ Watcher, err error) {
+func WatchChangeFeed(ctx context.Context, cx *sqlx.DB, tableName string, out EventChannel, cursor time.Time) (_ Watcher, err error) {
+	c := make(EventChannel)
+	deduplicate(c, out, 100)
+
 	var rows *sql.Rows
 	if cursor.IsZero() {
 		rows, err = cx.QueryContext(ctx, fmt.Sprintf("EXPERIMENTAL CHANGEFEED FOR %s RESOLVED = $1", tableName), heartBeatInterval.String())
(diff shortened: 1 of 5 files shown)


watcherx/changefeed.go
```diff
@@ -199,3 +202,62 @@ func WatchChangeFeed(ctx context.Context, cx *sqlx.DB, tableName string, c Event
...
+}
+
+func (r *ringBuffer) Contains(el fmt.Stringer) bool {
+	_, ok := r.seen[el.String()]
+	return ok
+}
+
+func (r *ringBuffer) Add(el fmt.Stringer) {
+	// Evict the oldest key.
+	if oldestKey := r.keys[r.keyIdx%r.capacity]; oldestKey != "" {
```
aeneasr, line 255: This ring buffer is not thread-safe, is there not some stdlib tooling we can use instead?

watcherx/changefeed.go
```diff
@@ -68,7 +68,10 @@ const heartBeatInterval = time.Second
 // Watcher.DispatchNow() does not have an effect in this method.
 //
 // This watcher is blocking to allow proper context cancellation and clean up.
-func WatchChangeFeed(ctx context.Context, cx *sqlx.DB, tableName string, c EventChannel, cursor time.Time) (_ Watcher, err error) {
+func WatchChangeFeed(ctx context.Context, cx *sqlx.DB, tableName string, out EventChannel, cursor time.Time) (_ Watcher, err error) {
+	c := make(EventChannel)
+	deduplicate(c, out, 100)
```
aeneasr, line 73: How do you determine this (the 100 mark)? Would it be better to instead set a time (e.g. 15 seconds)?

watcherx/changefeed.go
hperl: It's not built for concurrent use, yes. The usage is just within the `deduplicate` function.

watcherx/changefeed.go
aeneasr: Ristretto is an eventually consistent cache that has no guarantee that writes end up in the cache! For de-duplication, we would indeed need something more consistent like the ring buffer used here. There are projects like https://github.com/patrickmn/go-cache but we should try to keep the memory consumption of the cache, even on high load, to < 50mb

watcherx/changefeed.go
aeneasr: That makes sense, in that case I think a short codedoc comment at the ringbuffer struct / instantiator or methods (Add) would be helpful for future generations

hperl: OK, will add!

watcherx/changefeed.go
hperl: The idea behind the ring buffer with the fixed number of elements was that we want constant runtime (as it sits between the changefeed and the consumers) and constant memory consumption. If we evict not based on the total elements in the set/cache, but based on time, the data structure must support an arbitrary amount of elements.

I still think the impementation with the constant number of elements is fine, especially since the previous behaviour was that there was no deduplication at all. If a time-based eviction policy is desired, the eviction probably needs to happen in a separate goroutine to not block forwarding the events. In any case I'd rather implement this for this special use case than use a general-purpose datastructure.

aeneasr: SGTM

watcherx/changefeed.go
```diff
@@ -199,3 +203,75 @@ func WatchChangeFeed(ctx context.Context, cx *sqlx.DB, tableName string, c Event
...
+}
+
+func (r *ringBuffer) Contains(el fmt.Stringer) bool {
+	_, ok := r.seen[r.key(el)]
+	return ok
+}
+
+func (r *ringBuffer) Add(el fmt.Stringer) {
+	// Evict the oldest key.
+	if oldestKey := r.keys[r.keyIdx%r.capacity]; oldestKey != 0 {
```
aeneasr, line 268: Can you please add a test for eviction?

hperl: Test for eviction is in https://github.com/hperl/x/blob/92477b79aca33bec98145f6704e7507156ce58d3/watcherx/changefeed_test.go#L209.

Btw, I changed the implementation to use a hash function (`hash/maphash`) in the ring buffer, however I don't deal with collisions. Is this sensible / a trade worth taking?

aeneasr: It's important that things stay consistent here, it's better to send an event twice than it is to lose an event due to this.

aeneasr: Yeah I think it makes sense to use sha1 or something here

hperl: Done!

aeneasr:
@hperl ready for review?

hperl:
@aeneasr yes, ready :+1:

aeneasr (approved):

PR #555, opened 2022-06-27
Add http client cmdx helper

aeneasr:
@Benehiko PTAL https://github.com/ory/x/commit/e294636fb31bbe24969ffc42b5e17c3807b52783

We should use this CORS handler for backoffice, kratos and remove the CORS handling from the bo proxy so that oryapis.com can also have CORS


Benehiko:
Will take a look :)

PR #573, opened 2022-07-25
fix(tracing): enable trace propagation for OTEL

pschultz:
The middleware already attempts to extract a parent trace from the HTTP
request. However, that does nothing unless a global OTEL propagator is
configured. This patch adds that configuration.

An alternative is to configure the propagator only for the otel Bridge:

    bridge := otelOpentracing.NewBridgeTracer()
    bridge.SetTextMapPropagator(...)

But considering that the bridge only exists due to the use of a
deprecated package it seems more appropriate to configure the global
propagator in the non-deprecated package.

## Why it didn't work before

The tracing middleware calls `remoteContext, _ := opentracing.GlobalTracer().Extract(...)` to get the parent span, if any:

https://github.com/ory/x/blob/df7f8ab7bc01df3c1584b2bd9d2c443197ea1157/tracing/middleware.go#L23-L27

`opentracing.GlobalTracer()` returns the bridge set up earlier:

https://github.com/ory/x/blob/df7f8ab7bc01df3c1584b2bd9d2c443197ea1157/tracing/tracer.go#L205-L209

`*BridgeTracer.Extract` uses either its configured propagator, or the global one:

```go
	ctx := t.getPropagator().Extract(context.Background(), propagation.HeaderCarrier(header))
```

```go
func (t *BridgeTracer) getPropagator() propagation.TextMapPropagator {
	if t.propagator != nil {
		return t.propagator
	}
	return otel.GetTextMapPropagator()
}
```

https://github.com/open-telemetry/opentelemetry-go/blob/4bf6150fa94e18bdf01c96ed78ee6d1c76f8e308/bridge/opentracing/bridge.go#L666

https://github.com/open-telemetry/opentelemetry-go/blob/4bf6150fa94e18bdf01c96ed78ee6d1c76f8e308/bridge/opentracing/bridge.go#L678-L683

The default global propagator is a noop implementation:

```go
func defaultPropagatorsValue() *atomic.Value {
	v := &atomic.Value{}
	v.Store(propagatorsHolder{tm: newTextMapPropagator()})
	return v
}
```

```go
func newTextMapPropagator() *textMapPropagator {
	return &textMapPropagator{
		noop: propagation.NewCompositeTextMapPropagator(),
	}
}
```

https://github.com/open-telemetry/opentelemetry-go/blob/19294aab4cd8b52afc34b77403e7447109c9b858/internal/global/state.go#L94-L98

https://github.com/open-telemetry/opentelemetry-go/blob/19294aab4cd8b52afc34b77403e7447109c9b858/internal/global/propagator.go#L38-L42

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.
- [X] I confirm that this pull request does not address a security vulnerability. 
      I confirm that I got green light (please contact [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push the changes.
- [ ] I have added necessary documentation within the code base (if appropriate).

### DIFF
--- a/otelx/otlp.go
+++ b/otelx/otlp.go
@@ -6,6 +6,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
@@ -43,6 +44,11 @@ func SetupOTLP(t *Tracer, tracerName string) (trace.Tracer, error) {
 
 	tp := sdktrace.NewTracerProvider(tpOpts...)
 	otel.SetTracerProvider(tp)
+
+	otel.SetTextMapPropagator(propagation.NewCompositeTextMapPropagator(
+		propagation.TraceContext{},
+		propagation.Baggage{},
+	))
 
 	return tp.Tracer(tracerName), nil
 }
--- a/tracing/tracer.go
+++ b/tracing/tracer.go
@@ -27,6 +27,7 @@ import (
 	"go.opentelemetry.io/otel"
 	otelOpentracing "go.opentelemetry.io/otel/bridge/opentracing"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	otelSdkTrace "go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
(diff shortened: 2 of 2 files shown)


icyphox:
Hey there, thanks for this patch—looks good! We can merge it as is (after CI is green), but most of the Ory suite (barring Oathkeeper) have switched to the [otelx](https://github.com/ory/x/tree/master/otelx) package for tracing—just FYI.

pschultz:
So this change will not affect hydra? hydra at the current master depends on github.com/ory/x v0.0.368, and otelx doesn't exist in that version yet. That is why I wasn't aware the package even exists.

grep tells me that the only call to SetTextMapPropagator is in otelx.SetupJaeger, so presumably the otelx package has the same problem as before, but I have no (easy) way of verifying that because, as mentioned earlier, hydra doesn't use the otelx package.

Do you want me to add the call in otelx.SetupOTLP too?

icyphox:
> So this change will not affect hydra? hydra at the current master depends on github.com/ory/x v0.0.368, and otelx doesn't exist in that version yet. That is why I wasn't aware the package even exists.

Ah yes, apologies. The switch to otelx hasn't yet been released (it's on the v2.x branch). So your current change will indeed affect Hydra. 

> grep tells me that the only call to SetTextMapPropagator is in otelx.SetupJaeger, so presumably the otelx package has the same problem as before, but I have no (easy) way of verifying that because, as mentioned earlier, hydra doesn't use the otelx package.

Yeah, this is likely. I think I wrongly assumed that opentelemetry-go configures a default propagator (as TraceContext and Baggage). It's been explicitly configured in otelx.SetupJaeger because we make use of B3 and Jaeger propagators. 😅 
 
> Do you want me to add the call in otelx.SetupOTLP too?

Yes, if that isn't too much trouble. :) Thanks, and good catch.

pschultz:
I amended the commit and rebased it on master.

icyphox (approved):
Awesome, this is good to merge—paging @aeneasr.

aeneasr (approved):

aeneasr:
Awesome, thank you! 🎉 Your contribution makes Ory better :)

PR #602, opened 2022-09-20
feat: add trunc func

kelkarajay:
- Introduce func to return string after truncation by length specified

### DIFF
--- a/stringsx/truncate.go
+++ b/stringsx/truncate.go
@@ -4,14 +4,15 @@ import "unicode/utf8"
 
 // TruncateByteLen returns string truncated at the end with the length specified
 func TruncateByteLen(s string, length int) string {
-	if length > 0 && len(s) > length {
-		res := s[:length]
-
-		// in case we cut in the middle of an utf8 rune, we have to remove the last byte as well until it fits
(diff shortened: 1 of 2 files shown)


zepatrik (changes requested):
Just some minor comments :wink:

stringsx/truncate.go
```diff
@@ -0,0 +1,17 @@
...
+	if length > 0 && len(s) > length {
+		res := s[:length]
+
+		// in case we cut in the middle of an utf8 rune, we have to remove the last byte as well until it fits
+		for !utf8.ValidString(res) {
+			res = res[:len(res)-1]
+		}
+		return res
+	}
+	return s
```
zepatrik, line 16: Actually it would be better to invert the if here for code style. But just a minor detail really :wink: 
```suggestion
	if length <= 0 || len(s) <= length {
		return s
	}

	res := s[:length]

	// in case we cut in the middle of an utf8 rune, we have to remove the last byte as well until it fits
	for !utf8.ValidString(res) {
		res = res[:len(res)-1]
	}
	return res
```

zepatrik (approved):

PR #623, opened 2022-10-25
feat: keyset pag. sort direction option added with asc default

kelkarajay:


pagination/keysetpagination/paginator.go
```diff
@@ -41,6 +42,13 @@ func (p *Paginator) IsLast() bool {
 	return p.isLast
 }
 
+func (p *Paginator) SortDirection() string {
```
zepatrik, line 45: We cannot really change the sort direction, as we would have to flip the comparison as well. Let's just support ascending for now. Other use-cases are still possible by building the query yourself form the paginator properties.

pagination/keysetpagination/paginator.go
```diff
@@ -62,7 +70,8 @@ func Paginate[I Item](p *Paginator) pop.ScopeFunc {
 	return func(q *pop.Query) *pop.Query {
 		return q.
 			Limit(p.Size()+1).
-			Where(fmt.Sprintf("%s > ?", id), p.Token())
+			Where(fmt.Sprintf(`"%s" > ?`, id), p.Token()).
+			Order(fmt.Sprintf(`"%s" %s`, id, p.SortDirection()))
```
zepatrik, line 74:
```suggestion
			Where(fmt.Sprintf(`%q > ?`, id), p.Token()).
			Order(fmt.Sprintf(`%q ASC`, id))
```

This automatically adds double-quotes :wink:

zepatrik (commented):
Just some minor remarks :wink:

zepatrik (approved):

pagination/keysetpagination/paginator.go
```diff
@@ -62,7 +62,8 @@ func Paginate[I Item](p *Paginator) pop.ScopeFunc {
 	return func(q *pop.Query) *pop.Query {
 		return q.
 			Limit(p.Size()+1).
-			Where(fmt.Sprintf("%s > ?", id), p.Token())
+			Where(fmt.Sprintf(`%q > ?`, id), p.Token()).
+			Order(fmt.Sprintf(`%q ASC`, id))
```
aeneasr, line 66: This should be:

```
			Where(fmt.Sprintf("`%s` > ?", id), p.Token()).
			Order(fmt.Sprintf("`%s` ASC", id))
```

aeneasr: If it doesn't work across databases we will have to add a switch statement for the dialect! I think pop also has an escape helper if I'm not mistaken.

PR #625, opened 2022-10-26
feat: enable ANSI_QUOTES sql_mode to interpret " as identifiers

kelkarajay:
https://github.com/ory/x/pull/623 introduced a change to support using reserved keywords for pagination key column. This breaks queries on MySQL. 

The query parameter `sql_mode` is set to enable `ANSI_QUOTES` along with the defaults 

Reference to MySQL doc - https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes

### DIFF
--- a/sqlcon/parse_opts.go
+++ b/sqlcon/parse_opts.go
@@ -105,6 +105,7 @@ func FinalizeDSN(l *logrusx.Logger, dsn string) string {
 
 		q.Set("multiStatements", "true")
 		q.Set("parseTime", "true")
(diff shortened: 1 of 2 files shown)


aeneasr (changes requested):
I don't think it is a good idea to change the database mode all of Ory's SQL ecosystem because of one exception!

It should be possible to escape the reserved keyword in the paginator using https://github.com/ory/x/pull/623/files#r1006462207

Changing how we interpret SQL escaping is a risky operation with regards to potential SQL injection attack surface.

PR #626, opened 2022-10-27
fix: paginate escape column name

kelkarajay:
Paginate based on Dialect check for mysql

### DIFF
--- a/pagination/keysetpagination/paginator.go
+++ b/pagination/keysetpagination/paginator.go
@@ -60,17 +60,11 @@ func Paginate[I Item](p *Paginator) pop.ScopeFunc {
 	var item I
 	id := (&pop.Model{Value: item}).IDField()
 	return func(q *pop.Query) *pop.Query {
-		if q.Connection.Dialect.Name() == "mysql" {
-			return q.
-				Limit(p.Size()+1).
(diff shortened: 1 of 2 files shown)


kelkarajay:
https://github.com/ory/kratos/pull/2818 is now based off this version. Master seems to have changes that break private/loopback IP related tests on Kratos

pagination/keysetpagination/paginator.go
```diff
@@ -60,6 +60,13 @@ func Paginate[I Item](p *Paginator) pop.ScopeFunc {
 	var item I
 	id := (&pop.Model{Value: item}).IDField()
 	return func(q *pop.Query) *pop.Query {
+		if q.Connection.Dialect.Name() == "mysql" {
+			return q.
+				Limit(p.Size()+1).
+				Where(fmt.Sprintf("`%s` > ?", id), p.Token()).
+				Order(fmt.Sprintf("`%s` ASC", id))
```
aeneasr, line 67: I looked at pop, you should be able to use `q.Connection.Dialect.Quote()` which will quote the string according to the database being used, making sure this works on every dialect!

aeneasr (approved):
Perfect, thanks!

PR #630, opened 2022-11-03
feat: x/otelx prep work for tracing enhancements

alnr:
Ref https://github.com/ory-corp/cloud/issues/3008

Prep work for more changes in kratos/cloud.

Threw in a couple of lint issues my IDE kept bringing up.

### DIFF
--- a/otelx/jaeger.go
+++ b/otelx/jaeger.go
@@ -18,8 +18,8 @@ import (
 // Optionally, Config.Providers.Jaeger.LocalAgentAddress can be set.
 // NOTE: If Config.Providers.Jaeger.Sampling.ServerURL is not specfied,
 // AlwaysSample is used.
-func SetupJaeger(t *Tracer, tracerName string) (trace.Tracer, error) {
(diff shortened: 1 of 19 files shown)


otelx/jaeger.go
```diff
@@ -18,8 +18,8 @@ import (
 // Optionally, Config.Providers.Jaeger.LocalAgentAddress can be set.
 // NOTE: If Config.Providers.Jaeger.Sampling.ServerURL is not specfied,
 // AlwaysSample is used.
-func SetupJaeger(t *Tracer, tracerName string) (trace.Tracer, error) {
-	host, port, err := net.SplitHostPort(t.Config.Providers.Jaeger.LocalAgentAddress)
+func SetupJaeger(t *Tracer, tracerName string, c *Config) (trace.Tracer, error) {
```
Demonsthere, line 21: Does it make sense to pass 2 times almost the same object? Or can `t.Config` differ from `Config`?

alnr: I've removed the `Config` field from the tracer, so we need to pass the config manually.

Demonsthere (approved):

PR #640, opened 2022-11-25
feat: add additional column ordering to keysetpagination

jonas-jonas:
cc @zepatrik

## Checklist

### DIFF
--- a/pagination/keysetpagination/header.go
+++ b/pagination/keysetpagination/header.go
@@ -94,7 +94,11 @@ func Header(w http.ResponseWriter, u *url.URL, p *Paginator) {
 func Parse(q url.Values) ([]Option, error) {
 	var opts []Option
 	if q.Has("page_token") {
-		opts = append(opts, WithToken(q.Get("page_token")))
+		pageToken, err := url.QueryUnescape(q.Get("page_token"))
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		opts = append(opts, WithToken(pageToken))
 	}
 	if q.Has("page_size") {
 		size, err := strconv.Atoi(q.Get("page_size"))
--- a/pagination/keysetpagination/header.go
+++ b/pagination/keysetpagination/header.go
@@ -83,22 +83,26 @@ func header(u *url.URL, rel, token string, size int) string {
 // It contains links to the first and next page, if one exists.
 func Header(w http.ResponseWriter, u *url.URL, p *Paginator) {
 	size := p.Size()
-	w.Header().Set("Link", header(u, "first", p.defaultToken, size))
+	w.Header().Set("Link", header(u, "first", p.defaultToken.Encode(), size))
 
 	if !p.IsLast() {
-		w.Header().Add("Link", header(u, "next", p.Token(), size))
+		w.Header().Add("Link", header(u, "next", p.Token().Encode(), size))
 	}
 }
 
 // Parse returns the pagination options from the URL query.
-func Parse(q url.Values) ([]Option, error) {
+func Parse(q url.Values, p PageTokenConstructor) ([]Option, error) {
 	var opts []Option
 	if q.Has("page_token") {
 		pageToken, err := url.QueryUnescape(q.Get("page_token"))
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
-		opts = append(opts, WithToken(pageToken))
+		parsed, err := p(pageToken)
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		opts = append(opts, WithToken(parsed))
 	}
 	if q.Has("page_size") {
 		size, err := strconv.Atoi(q.Get("page_size"))
--- a/pagination/keysetpagination/header_test.go
+++ b/pagination/keysetpagination/header_test.go
@@ -1,0 +1,52 @@
+package keysetpagination
+
+import (
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/instana/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHeader(t *testing.T) {
+	p := &Paginator{
+		defaultToken: StringPageToken("default"),
+		token:        StringPageToken("next"),
+		size:         2,
+	}
+
+	u, err := url.Parse("http://ory.sh/")
+	require.NoError(t, err)
+
+	r := httptest.NewRecorder()
+
+	Header(r, u, p)
+
+	links := r.HeaderMap["Link"]
+	assert.Len(t, links, 2)
+	assert.Contains(t, links[0], "page_token=default")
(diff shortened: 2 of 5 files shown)


pagination/keysetpagination/paginator.go
```diff
@@ -61,12 +87,28 @@ func (p *Paginator) ToOptions() []Option {
...
-	id := (&pop.Model{Value: item}).IDField()
+	model := &pop.Model{Value: item}
+	id := model.IDField()
 	return func(q *pop.Query) *pop.Query {
 		eid := q.Connection.Dialect.Quote(id)
+
+		tokenParts := parseToken(id, p.Token())
+		idValue := tokenParts[id]
+		if column, ok := model.Columns().Cols[p.additionalColumn.name]; ok {
+			quoteName := q.Connection.Dialect.Quote(column.Name)
```
zepatrik, line 98: From a quick look at the code in pop, `(Dialect).Quote` does not prevent SQL injection. We should clarify in the documentation for the function that no untrusted input should be passed down as the column name (or direction).
Do you have other ideas how we can prevent injection here? We could e.g. use reflection to verify that the string is a valid `db` struct tag, although that would be problematic with joins where you have to use dot notation to specify the table. But we would have the same problem with the ID field not being scoped in SQL.

zepatrik (commented):
Nice, this already looks very good :+1: 
Should we maybe allow multiple additional fields? We might not use it, but then it will be easy to do once it comes up.

pagination/keysetpagination/paginator.go
jonas-jonas: The line above this prevents SQL injection, as it checks whether, the supplied field is part of the model. If it's not, we ignore it and go to the `else` clause. IMO that's enough, but feel free to exploit ;)

pagination/keysetpagination/paginator.go
```diff
@@ -40,6 +47,25 @@ func (p *Paginator) Size() int {
...
+	}
+
+	r := map[string]string{}
+
+	for _, p := range tokens {
+		parts := strings.Split(p, "=")
+		if len(parts) != 2 {
+			continue
+		}
+		r[parts[0]] = parts[1]
```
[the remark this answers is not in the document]
jonas-jonas: oh nice! thanks.

pagination/keysetpagination/paginator.go
zepatrik: Ah ok, that's great then.

jonas-jonas:
@zepatrik I'd like to postpone the support for multiple fields for now. I've spent too much time on this already, and that would just complicate it, IMO. I'll add some more tests, but then it should be good to go.

pagination/keysetpagination/paginator.go
```diff
@@ -127,6 +210,16 @@ func WithSize(size int) Option {
 	}
 }
 
+func WithColumn(name string, order Order) Option {
```
zepatrik, line 213: Do you plan on exposing the order as a parameter to the HTTP client? In that case I think it would be better to not make it part of this option, but rather make it part of the token.
I mean, the token is basically a pointer to a row in a table of certain order. Therefore the order should also be derivable from the token correct?

zepatrik: Hm, in general it might not make sense to have this as an option. What happens when I first make a request to `/identities?page_size=100&additional_column=created_at,DESC`, but then the subsequent request is `/identities?page_size=100&page_token=.......`?

jonas-jonas: No, not yet. I think we need a more sophisticated method of transporting that information from the client to the library, than just delimited strings. I have some ideas, but I need to finish the courier lists first. After that, we can have another hack at this.

zepatrik: We just have to mind breaking changes, and should not redo it in a different way later on...

pagination/keysetpagination/page_token.go
```diff
@@ -0,0 +1,72 @@
...
+type MapPageToken map[string]string
+
+func (m MapPageToken) Parse(_ string) map[string]string {
+	return map[string]string(m)
+}
+
+const pageTokenColumnDelim = "/"
+
+func (m MapPageToken) Encode() string {
+	elems := []string{}
```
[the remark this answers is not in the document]
jonas-jonas: (y) performance . improved . :D

pagination/keysetpagination/page_token.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	}
+
+	return MapPageToken(r), nil
+}
+
+var _ PageTokenConstructor = NewMapPageToken
+var _ PageTokenConstructor = NewStringPageToken
+
+type PageTokenConstructor func(string) (PageToken, error)
```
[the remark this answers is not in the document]
jonas-jonas: Ah, nice. Don't think this changes much in the current PR. Or did you have something specific in mind?

pagination/keysetpagination/header_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	u, err := url.Parse("http://ory.sh/")
+	require.NoError(t, err)
+
+	r := httptest.NewRecorder()
+
+	Header(r, u, p)
+
+	links := r.HeaderMap["Link"]
+	assert.Len(t, links, 2)
```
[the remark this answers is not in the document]
jonas-jonas: Good catch, missed it this time^^

pagination/keysetpagination/header.go
```diff
@@ -83,18 +83,26 @@ func header(u *url.URL, rel, token string, size int) string {
...
 
 	if !p.IsLast() {
-		w.Header().Add("Link", header(u, "next", p.Token(), size))
+		w.Header().Add("Link", header(u, "next", p.Token().Encode(), size))
 	}
 }
 
 // Parse returns the pagination options from the URL query.
-func Parse(q url.Values) ([]Option, error) {
+func Parse(q url.Values, p PageTokenConstructor) ([]Option, error) {
```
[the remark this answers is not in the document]
jonas-jonas: Yeah, I was thinking the same thing, but in the end didn't want to bloat the token with more, though we could easily do `t:String` and `t:Map` or something, to keep it light.

But since, we now control the `encoding` and `parsing` we could change this with relatively little effort in the future.

pagination/keysetpagination/header_test.go
```diff
@@ -0,0 +1,55 @@
...
+	assert.Contains(t, links[0], "page_token=default")
+	assert.Contains(t, links[1], "page_token=next")
+}
+
+func TestHeader_WithIsLast(t *testing.T) {
```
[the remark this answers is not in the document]
jonas-jonas: Yup, moved them together.

zepatrik (approved):
Nice :tada:

PR #641, opened 2022-11-28
feat: lighter SSRF migration

hperl:
## Checklist

### DIFF
--- a/httpx/private_ip_validator.go
+++ b/httpx/private_ip_validator.go
@@ -8,8 +8,8 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-
-	"github.com/ory/x/stringsx"
+	"syscall"
+	"time"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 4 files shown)


httpx/private_ip_validator.go
```diff
@@ -80,29 +80,54 @@ var _ http.RoundTripper = (*NoInternalIPRoundTripper)(nil)
...
-		return nil, err
-	}
+	return NoInternalTransport.RoundTrip(request)
+}
+
+var NoInternalDialer = &net.Dialer{
+	Timeout:   30 * time.Second,
+	KeepAlive: 30 * time.Second,
+	Control: func(network, address string, _ syscall.RawConn) error {
+		if !(network == "tcp4" || network == "tcp6") {
```
aeneasr, line 103: From docs:

````
	// If Control is not nil, it is called after creating the network
	// connection but before actually dialing.
	//
	// Network and address parameters passed to Control method are not
	// necessarily the ones passed to Dial. For example, passing "tcp" to Dial
	// will cause the Control function to be called with "tcp4" or "tcp6".
```

Not sure if that means that the address could also be different?

Another question is: how does this deal with sockets? They should also not be allowed.

aeneasr (commented):
Very nice clean-up

httpx/private_ip_validator.go
hperl: The `address` will contain the resolved address, so if you `Dial("tcp", "google.com")`, the `Control callback will receive the resolved IP address, and most likely `tcp6` as the protocol. We check that we get only `tcp4` or `tcp6` and that the address is really an IP address (we don't do any lookup any more, and err if it is not a parseable IP address).

Regarding your other question, I expanded the test coverage, see: https://github.com/ory/x/commit/33e13f6b3a01b9b6b88c6348909bddd251d65c39#diff-ece808bf9f537051020f575d16079a3f53528a677d892509b133015eed4741dfR87-R125. In short, only TCP traffic is allowed, nothing else.

PR #652, opened 2023-01-03
fix: clean up instrumentedsql

alnr:
This gives much more details into what's happening in the SQL driver, and SQL query+argument span tags appear at the correct place:

image
image


aeneasr (approved):
Fixing a bunch of bugs by deleting code? Love it :)

aeneasr:
Once the formatter is happy we can merge!

PR #659, opened 2023-01-10
feat: keep light validation

zepatrik:


httpx/private_ip_validator.go
```diff
@@ -99,8 +107,21 @@ func (n NoInternalIPRoundTripper) RoundTrip(request *http.Request) (*http.Respon
...
+	}
+	if port == "" {
+		port = portMap[incoming.Scheme]
+	}
+
+	// Here we dial the upstream using our custom control before we to the actual round trip. This ensures only
+	// valid requests are made in the end.
+	_, err = NoInternalDialer.DialContext(request.Context(), "tcp", net.JoinHostPort(host, port))
+	if err != nil {
+		return nil, err
```
aeneasr, line 124: As far as I can read this, this will now open two connections (this connection is not closed btw) per request to "test" whether the IP falls in the range or not. What is the advantage over resolving the host to an IP and checking the IP instead? We're essentially now doing:

1. Look up IP
2. Open TCP
3. Close TCP
4. Look up IP (in default dialer)
5. Open TCP
6. Send request
7. Close TCP

With the original code:

1. Lookup IP
2. Validate IP
3. Use default handling

zepatrik: From my understanding existing connections are reused, but I have to check that again.

PR #661, opened 2023-01-13
fix: spurious spans from config reloader

alnr:
I've disassociated the context used for tracing during config reloads from what's being set in `configx.WithContext`.

More generally, I feel we store too much stuff in contexts and have too much magic of injecting configuration. We also store contexts inside structures occasionally, which is an anti-pattern. I'd like to clean this up more after this PR.


aeneasr:
As far as I can tell this looks good. @zepatrik please double confirm though.

PR #662, opened 2023-01-13
Improve eager parallel

aeneasr:
## Checklist

### DIFF
--- a/sqlxx/expand.go
+++ b/sqlxx/expand.go
@@ -3,6 +3,15 @@
 
 package sqlxx
 
+import (
+	"sort"
+	"strings"
+
+	"github.com/gobuffalo/pop/v6"
+	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
+)
+
 // Expandable controls what fields to expand for projects.
 type Expandable string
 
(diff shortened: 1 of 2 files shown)


sqlxx/expand.go
```diff
@@ -32,3 +42,52 @@ func (e Expandables) Has(search Expandable) bool {
...
+func (e Expandables) load(ctx context.Context, c *pop.Connection, m interface{}, level int, maxLevel int) error {
+	var eg errgroup.Group
+	for i := range e {
+		item := e[i].String()
+		if len(strings.Split(item, ".")) == level {
+			continue
+		}
+
+		eg.Go(func() error {
+			// We need a copy of the connection which is only possible using `WithContext` because
```
alnr, line 60: What does „copy a connection“ mean, exactly?

aeneasr: It just copies the struct, not the SQL connection itself :)

aeneasr:
Unfortunately, this approach did not prove effective. The problem is that we call `Load()` on the same memory address (the model struct), which causes race conditions when executed in parallel. :(

PR #665, opened 2023-01-26
fix: remove x/tracing and reduce dependencies

alnr:


aeneasr:
Thank you! Unfortunately, removing directories from ory/x is always tricky. Sometimes we have dependencies that use both ory/x and an older version of e.g. ory/kratos. By removing the directory, we can no longer upgrade ory/x because it will cause with the older ory/kratos that still uses this package. It could work, but it could also mean endless dependency pain.

alnr:
> Thank you! Unfortunately, removing directories from ory/x is always tricky. Sometimes we have dependencies that use both ory/x and an older version of e.g. ory/kratos. By removing the directory, we can no longer upgrade ory/x because it will cause with the older ory/kratos that still uses this package. It could work, but it could also mean endless dependency pain.

I think I have most or all our projects cloned locally and I couldn't find a reference to `x/tracing` anywhere except in Hydra, where it's a trivial fix. WDYT?

alnr:
We gotta start somewhere, right? I've also opened this https://github.com/ory/hydra/pull/3424 (which we're gonna need sonner or later any way), which removes the last `x/tracing` remnants from Hydra.

aeneasr (approved):

PR #667, opened 2023-02-16
fix: correctly deal with premarshalled json objects

CaptainStandby:
## Checklist

### DIFF
--- a/decoderx/http.go
+++ b/decoderx/http.go
@@ -93,7 +93,7 @@ func HTTPFormDecoder() HTTPDecoderOption {
 	}
 }
 
-// HTTPJSONDecoder configures the HTTP decoder to only accept form-data
+// HTTPJSONDecoder configures the HTTP decoder to only accept JSON data
 // (application/json).
 func HTTPJSONDecoder() HTTPDecoderOption {
 	return func(o *httpDecoderOptions) {
@@ -335,19 +335,10 @@ func (t *HTTP) decodeJSONForm(r *http.Request, destination interface{}, o *httpD
 	}
 
 	values := url.Values{}
-	var notJSONForm bool
 	parsed.ForEach(func(k, v gjson.Result) bool {
-		if v.IsArray() || v.IsObject() {
-			notJSONForm = true
-			return false
-		}
 		values.Set(k.String(), v.String())
 		return true
 	})
-
-	if notJSONForm {
-		return t.decodeJSON(r, destination, o, true)
-	}
 
 	if o.queryAndBody {
 		_ = r.ParseForm()
@@ -518,7 +509,7 @@ func (t *HTTP) decodeURLValues(values url.Values, paths []jsonschemax.Path, o *h
 						continue
 					}
 
-					raw, err = sjson.SetBytes(raw, path.Name, v)
+					raw, err = sjson.SetRawBytes(raw, path.Name, []byte(v))
 				case []map[string]interface{}:
 					raw, err = sjson.SetBytes(raw, path.Name, values[key])
 				}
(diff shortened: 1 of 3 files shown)


decoderx/http.go
```diff
@@ -93,7 +93,7 @@ func HTTPFormDecoder() HTTPDecoderOption {
 	}
 }
 
-// HTTPJSONDecoder configures the HTTP decoder to only accept form-data
+// HTTPJSONDecoder configures the HTTP decoder to only accept JSON data
```
CaptainStandby, line 96: This seemed like a typo, so i fixed it.

decoderx/http.go
```diff
@@ -335,20 +335,11 @@ func (t *HTTP) decodeJSONForm(r *http.Request, destination interface{}, o *httpD
...
-			notJSONForm = true
-			return false
-		}
 		values.Set(k.String(), v.String())
 		return true
 	})
 
-	if notJSONForm {
-		return t.decodeJSON(r, destination, o, true)
-	}
```
CaptainStandby, line 350: This piece of code prevents a "mixed" scenario where part of the input JSON isn't flattened like so:
```json
{
  "dynamic_object": {
    "stuff": { "blub": [42, 3.14152], "fu": "bar" },
    "consent": true
  },
  "name.first": "Horst",
  "name.last": "Horstmann"
}
```
I removed the defaulting to `decodeJSON()` completely.
I added some tests around this and also no existing tests were failing, so I think we are good.

decoderx/http.go
```diff
@@ -518,7 +509,7 @@ func (t *HTTP) decodeURLValues(values url.Values, paths []jsonschemax.Path, o *h
 						continue
 					}
 
-					raw, err = sjson.SetBytes(raw, path.Name, v)
+					raw, err = sjson.SetRawBytes(raw, path.Name, []byte(v))
```
CaptainStandby, line 512: As we expect an object in this case, we can assume the value is premarshalled. The validation happening later will catch violations of this assumption.

decoderx/http.go
aeneasr: We have option `expectJSONFlattened` which is set by:

```
// HTTPDecoderJSONFollowsFormFormat if set tells the decoder that JSON follows the same conventions
// as the form decoder, meaning `{"foo.bar": "..."}` is translated to `{"foo": {"bar": "..."}}`.
func HTTPDecoderJSONFollowsFormFormat() HTTPDecoderOption {
	return func(o *httpDecoderOptions) {
		o.expectJSONFlattened = true
		o.keepRequestBody = true
	}
}
```

if that is set, we trigger `decodeJSONForm` but that in turn triggers regular JSON decoding if we encouter a complex type in the payload. This was introduced as part of ed839286 which mentions explicitly `fix(decoderx): improve form type detection`. I wasn't able to quickly find what caused this commit in Ory Kratos (probably some type of bug) but it would be good to understand the original reasoning behind this particular behavior (maybe it was a bug report for a specific edge case).

decoderx/http.go
aeneasr: That makes sense to me

aeneasr (approved):
Overall I think these changes make sense. You disable the wonky "if the type is not simple, then use a completely different decoding mechanism" and instead properly handle complex types in `SetRawBytes`. If you could find for me the commit / issue in Ory Kratos that relates to https://github.com/ory/x/commit/ed839286fe6f94f83958f0328a036a296c7d96d4 I think we can quickly move forward with this :)

By the way, do e2e tests pass if you use this version of Ory X in Ory Kratos?

CaptainStandby:
Kratos tests run without errors with this change. See: https://github.com/ory/kratos/pull/3113/checks

aeneasr:
Nice!

CaptainStandby:
The related change in Kratos is [here](https://github.com/ory/kratos/commit/d420fe6e8a491b20063d4bfeaa0a841058087d32) which was merged as part of [this PR](https://github.com/ory/kratos/pull/1306). Sadly there is not a lot of information on the specific problem that was solved back then.

PR #666, opened 2023-02-07
fix(watcherx): do not close channel to make it reusable for multiple watchers

zepatrik:

PR #671, opened 2023-02-20
fix(configx): goroutine leak

zepatrik:
#661 attempted to improve context usage in `configx` by using `context.Background()` instead of storing context in the struct.
With the changes from #666, `watcherx` however does not automatically close channels, so `configx` started leaking go routines.
We can actually easily avoid storing context, but also passing it down to the watcher such that context cancelation automatically trickles down, no need for a `.Close()` function.


alnr:
I find the use of contexts for cancelation of background workers a little bit wonky. Constructors should generally not take contexts as arguments, in my opinion. That's not a thing in the Go standard library either. Contexts are for request-scoped data.

I'm not gonna block this PR on those grounds. But we definitely need some docs around these functions, clearly explaining what they use the context for and how and when any background goroutines are reclaimed.

alnr:
I think a better API for the config provider thingy would be to not spawn background goroutines directly in the constructor. Instead, have a dedicated opt-in function [for example `func (*Provider) BackgroundReload() (stop func())`], which makes it obvious where the goroutines are coming from and makes it harder to leak them.

PR #674, opened 2023-03-02
feat: collect Prometheus metrics on automatic transaction retries (cockroachdb)

alnr:
This should allow us to monitor transaction contention in Grafana.

### DIFF
--- a/popx/transaction.go
+++ b/popx/transaction.go
@@ -5,11 +5,12 @@ package popx
 
 import (
 	"context"
+	"runtime"
 
 	"github.com/cockroachdb/cockroach-go/v2/crdb"
-	"github.com/jmoiron/sqlx"
-
 	"github.com/gobuffalo/pop/v6"
+	"github.com/jmoiron/sqlx"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 type transactionContextKey int
@@ -35,7 +36,13 @@ func Transaction(ctx context.Context, connection *pop.Connection, callback func(
 				return err
 			}
 
+			attempt := 0
 			return crdb.ExecuteInTx(ctx, sqlxTxAdapter{transaction.TX.Tx}, func() error {
+				attempt++
+				if attempt > 1 {
+					caller := caller()
+					TransactionRetries.WithLabelValues(caller).Inc()
+				}
 				return callback(WithTransaction(ctx, transaction), transaction)
 			})
 		})
@@ -73,4 +80,36 @@ func (s sqlxTxAdapter) Commit(ctx context.Context) error {
 
 func (s sqlxTxAdapter) Rollback(ctx context.Context) error {
 	return s.Tx.Rollback()
+}
+
+var (
+	TransactionRetries = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "ory_x_popx_cockroach_transaction_retries_total",
+		Help: "Counts the number of automatic CockroachDB transaction retries",
+	}, []string{"caller"})
+	_             = TransactionRetries.WithLabelValues(unknownCaller) // make sure the metric is always present
+	unknownCaller = "unknown"
+)
+
+func caller() string {
+	pc := make([]uintptr, 3)
+	// The number stack frames to skip was determined by putting a breakpoint in
+	// ory/kratos and looking for the topmost frame which isn't from ory/x or
+	// gobuffalo/pop.
+	n := runtime.Callers(8, pc)
+	if n == 0 {
+		return unknownCaller
+	}
+	pc = pc[:n]
+	frames := runtime.CallersFrames(pc)
+	for {
+		frame, more := frames.Next()
+		if frame.Function != "" {
+			return frame.Function
+		}
+		if !more {
+			break
+		}
+	}
+	return unknownCaller
 }
--- a/popx/transaction.go
+++ b/popx/transaction.go
@@ -41,7 +41,7 @@ func Transaction(ctx context.Context, connection *pop.Connection, callback func(
 				attempt++
 				if attempt > 1 {
 					caller := caller()
-					TransactionRetries.WithLabelValues(caller).Inc()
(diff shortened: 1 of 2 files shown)


jonas-jonas (approved):
LGTM.

popx/transaction.go
```diff
@@ -35,7 +36,13 @@ func Transaction(ctx context.Context, connection *pop.Connection, callback func(
 				return err
 			}
 
+			attempt := 0
 			return crdb.ExecuteInTx(ctx, sqlxTxAdapter{transaction.TX.Tx}, func() error {
+				attempt++
+				if attempt > 1 {
+					caller := caller()
+					TransactionRetries.WithLabelValues(caller).Inc()
```
zepatrik, line 44: Does this mean that the counters are per caller and not per calls? So two retries for one tx, or one retry each for two txs, all from the same caller, result in a retry counter of two? Is that desired?

popx/transaction_test.go
```diff
@@ -130,3 +143,28 @@ func (t popWriteSkewTest) UpdateBalance(
...
+		done    = make(chan struct{})
+	)
+	go func() {
+		defer close(done)
+		for m := range mChan {
+			metrics = append(metrics, m)
+		}
+	}()
+	TransactionRetries.Collect(mChan)
+	close(mChan)
```
zepatrik, line 160: Closing a channel as the receiver? This looks fishy :sweat_smile:
Why not just range over the channel in this function instead of the go routine? Also this seems racy that you close the channel while `.Collect` might still write to it :thinking:

popx/transaction_test.go
```diff
@@ -130,3 +143,28 @@ func (t popWriteSkewTest) UpdateBalance(
...
+	<-done
+	for _, m := range metrics {
+		var pb dto.Metric
+		require.NoError(t, m.Write(&pb))
+		require.NotNil(t, pb.Counter)
+		require.NotEmpty(t, pb.Label)
+		return *pb.Label[0].Name, *pb.Label[0].Value, int(*pb.Counter.Value)
```
zepatrik, line 167: Why return in the loop? If you are only interested in the first value you can just make this whole function A LOT easier. Also, in case `len(metrics) == 0` we would just silently return the zero values, but I guess the test should fail.

zepatrik (changes requested):

popx/transaction.go
```diff
@@ -74,3 +81,35 @@ func (s sqlxTxAdapter) Commit(ctx context.Context) error {
...
+var (
+	TransactionRetries = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "ory_x_popx_cockroach_transaction_retries_total",
+		Help: "Counts the number of automatic CockroachDB transaction retries",
+	}, []string{"caller"})
+	_             = TransactionRetries.WithLabelValues(unknownCaller) // make sure the metric is always present
+	unknownCaller = "unknown"
+)
+
+func caller() string {
```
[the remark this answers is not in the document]
alnr: There's really no way to tell if the number of skipped frames is "correct", since usage will vary. 8 is my best guess from common usage in kratos. We'll have to try and see in actual usage how many suboptimal labels we get.

popx/transaction.go
```diff
@@ -74,3 +81,35 @@ func (s sqlxTxAdapter) Commit(ctx context.Context) error {
 func (s sqlxTxAdapter) Rollback(ctx context.Context) error {
 	return s.Tx.Rollback()
 }
+
+var (
+	TransactionRetries = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "ory_x_popx_cockroach_transaction_retries_total",
+		Help: "Counts the number of automatic CockroachDB transaction retries",
+	}, []string{"caller"})
```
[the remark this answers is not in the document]
alnr: `TransactionRetries` is global, but internally synchronized. It is not registered, but is expected to be registered for collection by the user in their app. Kratos etc seem to use the prometheus built-in global registry. The code would be something like:

```
prometheus.Register(popx.TransactionRetries)
```

popx/transaction.go
```diff
@@ -74,3 +81,35 @@ func (s sqlxTxAdapter) Commit(ctx context.Context) error {
...
+	n := runtime.Callers(8, pc)
+	if n == 0 {
+		return unknownCaller
+	}
+	pc = pc[:n]
+	frames := runtime.CallersFrames(pc)
+	for {
+		frame, more := frames.Next()
+		if frame.Function != "" {
+			return frame.Function
```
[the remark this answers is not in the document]
alnr: Package name is part of the function name. Most (or all?) usages of `popx.Transaction` are used exactly once inside a function, so I think disambiguation is not necessary.

popx/transaction.go
alnr: Yes, the counter is per caller. Otherwise you'd get a cardinality explosion in the labels (very common problem in Prometheus-style metrics). The idea is to have a dashboard graphing the number of transaction retries, grouped by the calling function (think `persister.CreateIdentity`, `session.Create` etc). That allows you to identify and monitor problematic code paths and fix them one-by-one.

I don't think having per-call events for retries is particularly helpful. They're a pretty normal occurence every once in a while, and aggregating them by caller gives us all the info we need without having to log. Logging would require injecting a logger, etc etc.

popx/transaction_test.go
alnr: This is lifted, with modifications, from the prometheus/testutil package. It works, so 🤷 😁

popx/transaction_test.go
alnr: Yeah, I got carried away a little while modifying the code from prometheus/testutil. Let's just say testing prometheus metrics is not exactly straightforward.

I can revisit the code if you want. Then again 🤷 

LMKWYT.

zepatrik (approved):
Thanks for the explanations :+1:

PR #676, opened 2023-03-28
feat: add semantic conventions

CaptainStandby:
## Checklist

- [x] I have added the necessary documentation within the code base (if appropriate).

### DIFF
--- a/otelx/semconv/context.go
+++ b/otelx/semconv/context.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"net/http"
 
-	"github.com/urfave/negroni"
 	"go.opentelemetry.io/otel/attribute"
 
 	"github.com/ory/x/httpx"
@@ -36,7 +35,7 @@ func AttributesFromContext(ctx context.Context) []attribute.KeyValue {
 	return attrs
 }
 
-var Middleware negroni.HandlerFunc = func(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+func Middleware(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
 	next(rw, r.WithContext(ContextWithAttributes(r.Context(), AttrClientIP(httpx.ClientIP(r)))))
 }
 
--- a/otelx/semconv/context.go
+++ b/otelx/semconv/context.go
@@ -1,0 +1,47 @@
+// Copyright © 2023 Ory Corp
+// SPDX-License-Identifier: Apache-2.0
+
+package semconv
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/urfave/negroni"
+	"go.opentelemetry.io/otel/attribute"
+
+	"github.com/ory/x/httpx"
+)
+
+type contextKey int
+
+const contextKeyAttributes contextKey = iota
+
+func ContextWithAttributes(ctx context.Context, attrs ...attribute.KeyValue) context.Context {
+	existing, _ := ctx.Value(contextKeyAttributes).([]attribute.KeyValue)
+	return context.WithValue(ctx, contextKeyAttributes, append(existing, attrs...))
+}
+
+func AttributesFromContext(ctx context.Context) []attribute.KeyValue {
+	fromCtx, _ := ctx.Value(contextKeyAttributes).([]attribute.KeyValue)
+	uniq := make(map[attribute.Key]struct{})
+	attrs := make([]attribute.KeyValue, 0)
+	for i := len(fromCtx) - 1; i >= 0; i-- {
+		if _, ok := uniq[fromCtx[i].Key]; !ok {
+			uniq[fromCtx[i].Key] = struct{}{}
+			attrs = append(attrs, fromCtx[i])
+		}
+	}
+	reverse(attrs)
+	return attrs
+}
+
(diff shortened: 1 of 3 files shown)


CaptainStandby:
@alnr I think this would be a nice addition

otelx/semconv/context.go
```diff
@@ -0,0 +1,47 @@
...
+
+var Middleware negroni.HandlerFunc = func(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	next(rw, r.WithContext(ContextWithAttributes(r.Context(), AttrClientIP(httpx.ClientIP(r)))))
+}
+
+func reverse[S ~[]E, E any](s S) {
+	for i, j := 0, len(s)-1; i < j; i, j = i+1, j-1 {
+		s[i], s[j] = s[j], s[i]
+	}
+}
```
jonas-jonas, line 47: gotta love golang sometimes...

otelx/semconv/context.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/ory/x/httpx"
+)
+
+type contextKey int
+
+const contextKeyAttributes contextKey = iota
+
+func ContextWithAttributes(ctx context.Context, attrs ...attribute.KeyValue) context.Context {
+	existing, _ := ctx.Value(contextKeyAttributes).([]attribute.KeyValue)
```
jonas-jonas, line 21: Is it safe to ignore the second value here? What happens if the type assertion fails?

jonas-jonas (commented):
LGTM in general.

otelx/semconv/context.go
alnr: We use the `a, _ := ` syntax here to prevent the type assertion from panicking if the key does not exist in the context. Instead of panicking, the assertion will return `nil, false`. But we don't care about the `false`, since we proceed in the same manner anyway, i.e., appending to a (possibly nil) slice in the next step.

otelx/semconv/events.go
```diff
@@ -4,13 +4,57 @@
...
+	"github.com/gofrs/uuid"
+	otelattr "go.opentelemetry.io/otel/attribute"
+)
+
+type Event string
+
 const (
-	EventSessionIssued   = "SessionIssued"
-	EventIdentityCreated = "IdentityCreated"
+	EventSessionIssued     Event = "SessionIssued"
```
alnr, line 15: Couple things:

1. Do we really need all these events at the moment? Things like `RecoveryStarted` feel very fine-grained. Is that really a business-level event? Maybe we do, but adding a bunch of events without really thinking about their usefulness is not the point :)
2. The existing events `EventSessionIssued` and `EventIdentityCreated` have the `Event...` prefix in their names. We should either have this convention for all or none of the event names.
3. Please refer to [this doc](https://intranet.ory.sh/docs/dev/analytics/analytics-events/) to find out about the naming convention, especially the `ObjectAction` (with `Action` in past tense) convention. See references in the above link for rationales.
4. Please make sure that the event names correspond to what we normally call a business event in the Oryverse, and that it is understandable without too much context. Example: `SignupFailed`: in kratos we call this a "registration", not a signup. Example 2: `AccountLinked`. This is some OIDC-related concept, correct? It should make that clear somehow, since just `AccountLinked` is very generic.

CaptainStandby: One other thing I have been thinking about:
Maybe it would make sense to move these event definitions to the concrete projects they belong to (`EventSessionIssued` belongs to Kratos, for example)
Then we would only keep the types and helper function in here and won't have to update this repo when adding more events.

otelx/semconv/context.go
jonas-jonas: I see, thanks for the clarification.

otelx/semconv/context.go
```diff
@@ -0,0 +1,47 @@
...
+			attrs = append(attrs, fromCtx[i])
+		}
+	}
+	reverse(attrs)
+	return attrs
+}
+
+var Middleware negroni.HandlerFunc = func(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	next(rw, r.WithContext(ContextWithAttributes(r.Context(), AttrClientIP(httpx.ClientIP(r)))))
+}
```
CaptainStandby, line 41: Is there a reason not to write this as a function?
```suggestion
func Middleware(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
	next(rw, r.WithContext(ContextWithAttributes(r.Context(), AttrClientIP(httpx.ClientIP(r)))))
}
```

alnr: Yep, that's a useful suggestion. It would decouple this package from urfave/negroni (somewhat).

otelx/semconv/events.go
CaptainStandby: I have removed the event definitions from this package

jonas-jonas (approved):

alnr (approved):

alnr:
Please add the event you deleted here back into kratos after merging this :)

CaptainStandby:
> Please add the event you deleted here back into kratos after merging this :)

That's the plan :-)

PR #685, opened 2023-04-12
fix: health/alive handler never errors

alnr:
This is problematic during code generation in downstream projects ever since https://github.com/ory/herodot/pull/106. Since the handler never returns an error (a failing liveness check means you don't get a response at all), the simplest and most correct thing to do is to just not declare an error type at all.

### DIFF
--- a/healthx/handler.go
+++ b/healthx/handler.go
@@ -123,7 +123,6 @@ func (h *Handler) SetVersionRoutes(r router, opts ...Options) {
 //
 //	Responses:
 //	  200: healthStatus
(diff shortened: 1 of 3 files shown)


zepatrik (commented):
The handler itself might not error, but you could e.g. get a 504 or similar. I think you can provide `default` instead of a status code as well.

alnr:
> The handler itself might not error, but you could e.g. get a 504 or similar. I think you can provide `default` instead of a status code as well.

updated

PR #686, opened 2023-04-13
fix(decoderx): minor fixes

zepatrik:
Removed `io/ioutil` and other minor fixes as annotated.

Closes https://github.com/ory-corp/cloud/issues/4298 (once bumped upstream).

### DIFF
--- a/decoderx/http.go
+++ b/decoderx/http.go
@@ -10,7 +10,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


decoderx/http.go
```diff
@@ -326,7 +325,7 @@ func (t *HTTP) decodeJSONForm(r *http.Request, destination interface{}, o *httpD
 
 	var interim json.RawMessage
 	if err := json.NewDecoder(reader).Decode(&interim); err != nil {
-		return err
+		return errors.WithStack(herodot.ErrBadRequest.WithError(err.Error()).WithReason("Unable to decode form as JSON."))
```
zepatrik, line 328: Return `ErrBadRequest` here, so it is not a 500.

decoderx/http.go
```diff
@@ -539,27 +538,25 @@ func (t *HTTP) decodeURLValues(values url.Values, paths []jsonschemax.Path, o *h
 	return raw, nil
 }
 
-func (t *HTTP) decodeJSON(r *http.Request, destination interface{}, o *httpDecoderOptions, isRetry bool) error {
+func (t *HTTP) decodeJSON(r *http.Request, destination interface{}, o *httpDecoderOptions) error {
```
zepatrik, line 541: `isRetry` was always false, so I removed it accordingly.

aeneasr (approved):

PR #689, opened 2023-05-04
fix(otelx): memory leak in OpenTelemetry tracing middleware

jgraeger:
This PR addresses a memory leak issue in the OpenTelemetry tracing middleware.

The existing middleware invokes NewHandler for each incoming request, which results in the allocation of a new oteltracehttp.Handler.
```
func TraceHandler(h http.Handler, opts ...otelhttp.Option) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		NewHandler(h, r.URL.Path, opts...).ServeHTTP(w, r)
	})
}
```
In our setup, this led to an increase of over 300MB per day per Hydra instance in allocated memory. Similar observations have been reported in open-telemetry/opentelemetry-go-contrib#2413.

This PR modifies the middleware to utilize a SpanNameFormatter function instead of passing the path to NewHandler. This preserves the existing functionality while eliminating the memory leak.

The issue has been reported in projects using the library (e.g., ory/hydra#3495).

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.
- [X] I have referenced an issue containing the design document if my change introduces a new feature.
- [X] I confirm that this pull request does not address a security vulnerability. 
- [X] I have added tests that prove my fix is effective or that my feature works.
- [X] I have added the necessary documentation within the code base (if appropriate).

## Further comments

- The otelhttptrace applies the passed options in sequential order. We prepend the SpanNameFormatter to allow users to override the SpanNameHandler if needed.

### DIFF
--- a/otelx/middleware.go
+++ b/otelx/middleware.go
@@ -40,7 +40,12 @@ func NewHandler(handler http.Handler, operation string, opts ...otelhttp.Option)
 
 // TraceHandler wraps otelx.NewHandler, passing the URL path as the span name.
 func TraceHandler(h http.Handler, opts ...otelhttp.Option) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		NewHandler(h, r.URL.Path, opts...).ServeHTTP(w, r)
-	})
+	// Use a span formatter to set the span name to the URL path, rather than passing in the operation to NewHandler.
+	// This allows us to use the same handler for multiple routes.
+	middlewareOpts := []otelhttp.Option{
+		otelhttp.WithSpanNameFormatter(func(operation string, r *http.Request) string {
+			return r.URL.Path
+		}),
+	}
+	return NewHandler(h, "", append(middlewareOpts, opts...)...)
 }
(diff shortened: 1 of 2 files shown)


jgraeger:
> How did you identify this?

Hydra serves a large number of requests in our setup. As stated in the PR description, the memory consumption of our pods exceeded 20 GB over the weekend, indicating a clear problem.

I set up a local environment with otel-collector and used a script to send a large number of requests (always for the same client/subject - no need for complex request generation). I dumped the memory profile. Unfortunately, I no longer have the file, but it showed a lot of  allocations related to the global structures the OTEL stacks write to when creating a new handler. My profile excluded the startup phase, so It didn't include the heavy but unrelated/unproblematic allocations from it. Deploying the fix to our environment confirmed that it the large memleak was gone.

After implementing the fix, I found what I've linked in the description. Seems like this mistake occured in other projects as well.

aeneasr:
Interesting - did you use the standard Go memory profiling? We tried identifying the memory leak using the same methodologies but due to a secondary context leak caused by this were not able to identify the culprit.

I reran the CI - let’s see if it psses

alnr (approved):

alnr:
You my friend are amazing!

PR #692, opened 2023-06-05
fix(jsonx): enable PATCHing an object's field when the object does not exist yet

IchordeDionysos:
This PR fixes https://github.com/ory/kratos/issues/3244, which fails to add a field to `metadata_public`/`metadata_admin` when nothing yet exists in that object / when the objects are null.

To achieve this we instruct json-patch to create the missing path.

I had to update to a different version of `json-patch` I hope it's fine 🥹

https://github.com/ory/kratos/issues/3244
https://ory-community.slack.com/archives/C012USDT5QQ/p1685968287885839

## Checklist

- [x] I have referenced an issue containing the design document if my change introduces a new feature.
- [x] I have added the necessary documentation within the code base (if appropriate).

## Further comments

It's still TBD whether we should introduce the change globally or only enable the new behavior via a dedicated option.

I'm happy to work on changing the behavior by introducing a dedicated option (if that's required).

However, I would imagine that other APIs may also benefit from the change?!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/cockroachdb/cockroach-go/v2 v2.2.16
 	github.com/dgraph-io/ristretto v0.1.1
 	github.com/docker/docker v20.10.24+incompatible
-	github.com/evanphx/json-patch v5.6.0+incompatible
+	github.com/evanphx/json-patch/v5 v5.6.0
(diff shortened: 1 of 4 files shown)


zepatrik (approved):
This should not cause bugs elsewhere, as until now one gets an error in this case.

PR #695, opened 2023-06-15
fix: various lint issues and remove tracing from config provider

alnr:
This tracing instrumentation in the config provider was never useful and polluted our tracing backend.

I've fixed many lint issues found by `staticcheck`. Some remain, but those were more onerous to fix.


sqlcon/dockertest/onexit.go
```diff
@@ -34,7 +34,7 @@ func (at *OnExit) Add(f func()) {
 	at.once.Do(func() {
 		go func() {
 			c := make(chan os.Signal, 1)
-			signal.Notify(c, syscall.SIGINT, syscall.SIGKILL, syscall.SIGTERM)
```
alnr, line 37: SIGKILL cannot be trapped.

josex/encoding.go
```diff
@@ -40,7 +40,6 @@ func (r Base64Reader) Read(p []byte) (n int, err error) {
 		// Strip =
 		case 0x3D:
 			n = i
-			break
```
alnr, line 43: This `break` is ineffective.

metricsx/middleware.go
```diff
@@ -58,7 +57,7 @@ func Hash(value string) string {
 	hash := sha256.New()
 	_, err := hash.Write([]byte(value))
 	if err != nil {
-		panic(fmt.Sprintf("unable to hash value"))
+		panic("unable to hash value")
```
jonas-jonas, line 60: Since this was here before, it's probably not an issue, but panicking in a (open source?) metrics component seems pretty dangerous.

It seems this is called on startup of the services, potentially leading to crash loops? Could we maybe improve this?

Do you have an opinion on it?

jonas-jonas (approved):

jonas-jonas (approved):

PR #696, opened 2023-06-16
feat: allow Go migrations outside of an SQL transaction

alnr:


popx/migrator.go
```diff
@@ -141,23 +143,43 @@ func (m *Migrator) UpTo(ctx context.Context, step int) (applied int, err error)
...
+				l.Warn("Migration has requested running outside a transaction. Proceed with caution.")
+				if err := mi.RunnerNoTx(mi, c); err != nil {
 					return err
 				}
 
 				// #nosec G201 - mtn is a system-wide const
-				if _, err = conn.TX.Exec(fmt.Sprintf("INSERT INTO %s (version) VALUES ('%s')", mtn, mi.Version)); err != nil {
-					return errors.Wrapf(err, "problem inserting migration version %s", mi.Version)
+				if err := c.RawQuery(fmt.Sprintf("INSERT INTO %s (version) VALUES (?)", mtn), mi.Version).Exec(); err != nil {
+					return errors.Wrapf(err, "problem inserting migration version %s. YOUR DATABASE MAY BE IN AN INCONSISTENT STATE! MANUAL INTERVENTION REQUIRED!", mi.Version)
```
aeneasr, line 178: This is acceptable for an SQL migrator. Why do we need this feature - as in what's the concrete use case?

aeneasr (changes requested):

popx/migrator.go
alnr: Many DDL changes we perform semi-regularly are, in fact, not atomic at all and cannot easily be rolled back by `ROLLBACK;`. Having them inside a transaction can actually make the problem worse!

Example from CockroachDB:

```sql
BEGIN; -- step 0
ALTER TABLE t add column crdb_region crdb_internal_region NULL DEFAULT default_to_database_primary_region(gateway_region())::crdb_internal_region; -- step 1
INSERT INTO schema_migration (...) VALUES (...); -- step 2
COMMIT; -- step 3
```

If CRDB deems the modification in step 1 illegal, it will report that only after step 3 has concluded 🤯 and the row has in fact been committed to `schema_migration`. We get this error:

```
		transaction committed but schema change aborted with error:
		(0A000): default_to_database_primary_region(): unimplemented:
		cannot evaluate scalar expressions containing region operations
		in this context (SQLSTATE XXA00)
```

That leaves your database in an indeterminate state and definitely requires manual intervention!

By contrast, if the above DDL change was *not* wrapped in a transaction, the offending statement would fail immediately, before the row would be inserted in `schema_migration`, and there would be no inconsistency!

CRDB also explicitly [documents such limitations](https://www.cockroachlabs.com/docs/stable/online-schema-changes.html#limitations). Until now, we've never encountered such an error. Probably out of sheer luck. But the schema changes necessary for multi-region will, under the right circumstances, trigger such errors.

The second scenario where opting out of the transaction is for bulk updates. See https://www.cockroachlabs.com/docs/stable/bulk-update-data.html#write-a-batch-update-loop. Having such a loop inside a transaction clearly totally counters the point of the loop in the first place.

In other words, DDL changes within transactions very often do not work the way we would like them to. It's a leaky abstraction, and we need the ability to opt out of the transaction behavior for some migrations.

aeneasr: Very interesting, thank you for the write up! Two concerns come to mind:

1. This change improves CockroachDB migration stability, but decreases PostgreSQL & MySQL migration stability. Do we accept that and if not, how can we address this?
2. The `INSERT INTO` should run in a retryable transaction.

aeneasr: > Are you referring to INSERT INTO schema_migration?

Yes! Sometimes CRDB will return a retry error for whatever reason, and we should retry adding the value into schema_migration

> Non-transactional DDL changes should be comparatively rare, and they are always hand-written Go migrations. The recommendation would be to use non-transactional migrations only for those databases where it's beneficial, and default to transactional DDLs otherwise.

That sounds good. How do I define whether a migration should run in a transaction or not?

alnr: CRDB [automatically retries](https://www.cockroachlabs.com/docs/v23.1/transactions#automatic-retries) single statements, so that should be good as is.

For the other databases, we'd have to manually list the retryable error conditions. I couldn't find a definitive list of which errors are retryable for [MySQL](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html), [SQLite](http://sqlite.com/matrix/rescode.html). Even [Postgres](https://www.postgresql.org/docs/current/mvcc-serialization-failure-handling.html) is not quite trivial. Feels like it's not worth it to over-engineer this.

---

You define a transactional migration like this:
```
		popx.Migration{
				Path:      "transactional",
				Version:   "1",
				Name:      "gomigration_tx",
				Direction: "up",
				Type:      "go",
				DBType:    "all", // or "cockroach"
				Runner: func(m popx.Migration, c *pop.Connection, tx *pop.Tx) error {
					return nil
				},
		}
```
and a non-transactional one like this:

```
		popx.Migration{
				Path:      "non-transactional",
				Version:   "1",
				Name:      "gomigration_notx",
				Direction: "up",
				Type:      "go",
				DBType:    "all",
				RunnerNoTx: func(m popx.Migration, c *pop.Connection) error {
					return nil
				},
		}
```

aeneasr: OK SGTM

aeneasr (approved):

popx/migration_box_gomigration_test.go
```diff
@@ -160,3 +161,134 @@ func TestGoMigrations(t *testing.T) {
...
+				Version:   "1",
+				Name:      "gomigration no transaction",
+				Direction: "up",
+				Type:      "go",
+				DBType:    "all",
+				RunnerNoTx: func(m popx.Migration, c *pop.Connection) error {
+					if _, err := c.Store.Exec("INSERT INTO tests (i, j) VALUES (?, ?)", i1, j1); err != nil {
+						return errors.WithStack(err)
+					}
+					close(up1)
```
hperl, line 250: This took me a while to understand. You are passing control back here to see if the changes take effect immediately, right? Could you not just execute a closure instead which connection is bound to the top-level `c`?

Define in the top level:
```go
assertValues := func(i int) {
  var j int64
  require.NoError(t, c.Store.Get(&j, "SELECT j FROM tests WHERE i = ?", i))
  assert.Equal(t, j1, j)
```

And then call it as part of the migrations.

hperl (changes requested):
LGTM, with one small idea on how to improve the test :)

popx/migration_box_gomigration_test.go
aeneasr: From my side it's ok to merge as is

alnr: Yeah, the closure would work too. Would be a little bit neater perhaps.

This works, so 🤷

PR #705, opened 2023-07-17
fix: use incoming TLS to detect originalScheme

aeneasr:
## Checklist

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -82,7 +82,7 @@ func rewriter(o *options) func(*httputil.ProxyRequest) {
 
 		if forwardedProto := r.In.Header.Get("X-Forwarded-Proto"); forwardedProto != "" {
 			c.originalScheme = forwardedProto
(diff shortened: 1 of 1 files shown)


zepatrik:
Ah I see, so this was the issue :facepalm:

aeneasr:
Unfortunately not, it was just something I saw while browsing the code. The problem is that removing the X-Forwarded headers makes sense if your proxy is exposed directly on the internet. You would do this in your Kubernetes ingress, for example. But if your proxy (which is for us the case) is already in an internal network, the headers have already been stripped by the ingress, and you actually want to propagate them.

Therefore, everything breaks if these headers are removed, because we sometimes rely on them to identify the correct hostname. The fix is to revert these changes completely, hide them behind a feature flag, or use a feature flag to enable X-Forwarded headers again. It's just difficult to find all the places where this is needed but thankfully we have good tests around redirects that catch all the issues.

PR #706, opened 2023-07-18
feat(proxy): flag to keep forwarded headers

aeneasr:
Adds a feature flag to revert regressions introduced by switching to proxy.Rewrite.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -17,7 +17,7 @@ import (
 
 type (
 	RespMiddleware func(resp *http.Response, config *HostConfig, body []byte) ([]byte, error)
-	ReqMiddleware  func(req *http.Request, config *HostConfig, body []byte) ([]byte, error)
+	ReqMiddleware  func(req *httputil.ProxyRequest, config *HostConfig, body []byte) ([]byte, error)
 	HostMapper     func(ctx context.Context, r *http.Request) (*HostConfig, error)
 	options        struct {
 		hostMapper      HostMapper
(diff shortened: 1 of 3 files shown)


proxy/proxy.go
```diff
@@ -96,16 +99,28 @@ func rewriter(o *options) func(*httputil.ProxyRequest) {
...
+				if v := r.In.Header.Get(h); v != "" {
+					r.Out.Header.Set(h, v)
+				}
+			}
+		}
+
 		c.setScheme(r)
 		c.setHost(r)
 
-		*r.Out = *r.Out.WithContext(context.WithValue(ctx, hostConfigKey, c))
```
aeneasr, line 108: This is already done by `getHostConfig`

proxy/proxy.go
```diff
@@ -96,16 +99,28 @@ func rewriter(o *options) func(*httputil.ProxyRequest) {
...
 
-		c, err := o.getHostConfig(r.Out)
+		ctx, c, err := o.getHostConfig(ctx, r.In)
 		if err != nil {
 			o.onReqError(r.Out, err)
 			return
 		}
 
+		if c.TrustForwardedHeaders {
+			headers := []string{
```
aeneasr, line 109: This fixes the regression

proxy/proxy.go
```diff
@@ -17,8 +17,8 @@ import (
 
 type (
 	RespMiddleware func(resp *http.Response, config *HostConfig, body []byte) ([]byte, error)
-	ReqMiddleware  func(req *http.Request, config *HostConfig, body []byte) ([]byte, error)
-	HostMapper     func(ctx context.Context, r *http.Request) (*HostConfig, error)
+	ReqMiddleware  func(req *httputil.ProxyRequest, config *HostConfig, body []byte) ([]byte, error)
+	HostMapper     func(ctx context.Context, r *http.Request) (context.Context, *HostConfig, error)
```
aeneasr, line 21: It's no longer possible to use the context properly because we don't know whether we get the context from `req.In` or `req.Out` now. So context needs to be explicit now.

PR #712, opened 2023-08-08
docs(pagination): improve parameter descriptions

zepatrik:


pagination/migrationpagination/header.go
```diff
@@ -22,11 +22,11 @@ type RequestParameters struct {
...
 	//
 	// For example, page 2 might not follow after page 1. And even if page 3 and 5 exist, but page 4 might not exist.
+	// The first page can be retrieved by omitting this parameter. Following page pointers will be returned in the
+	// `Link` header.
 	//
 	// required: false
 	// in: query
-	// default: 1
-	// min: 1
 	Page int `json:"page"`
```
zepatrik, line 30: Actually, we should make this a string instead of an integer.
```suggestion
	Page string `json:"page"`
```
It is a breaking change in the SDK, but we will have to do the breaking change anyway at some point. Should have been a string from the beginning...

aeneasr: Why should this be a string?

PR #713, opened 2023-08-08
feat(configx): allow exceptions in immutables

zepatrik:
This makes it possible to allow hot-reloading e.g. the CORS origins, while the rest of the serve key is not hot-reloadable.

### DIFF
--- a/corsx/middleware.go
+++ b/corsx/middleware.go
@@ -19,6 +19,8 @@ import (
 //	  panic("implement me")
 //	})
 //	// ...
+//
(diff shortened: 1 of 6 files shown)


corsx/middleware.go
```diff
@@ -19,6 +19,8 @@ import (
 //	  panic("implement me")
 //	})
 //	// ...
+//
+// Deprecated: because this is not really practical to use, you should use CheckOrigin as the cors.Options.AllowOriginRequestFunc instead.
```
zepatrik, line 23: The problem with this was that the config has to come from the config provider, but where you initialize the middleware you have no access to the config.

aeneasr (approved):

PR #718, opened 2023-08-17
chore(jsonnetsecure): add tracing

zepatrik:


jsonnetsecure/jsonnet_processvm.go
```diff
@@ -26,10 +28,19 @@ func NewProcessVM(opts *vmOptions) VM {
...
+	defer span.End()
+
 	// We retry the process creation, because it sometimes times out.
 	const processVMTimeout = 1 * time.Second
-	return backoff.RetryWithData(func() (string, error) {
-		ctx, cancel := context.WithTimeout(p.ctx, processVMTimeout)
+	return backoff.RetryWithData(func() (_ string, err error) {
+		defer func() {
+			if err != nil {
+				span.AddEvent("retry", trace.WithAttributes(attribute.String("error", err.Error())))
```
zepatrik, line 39: Not sure if event is the right thing here, but seems to make sense. I can also add `err_0`, `err_1`, ... as attributes for each iteration.

jsonnetsecure/jsonnet_processvm.go
```diff
@@ -26,10 +30,17 @@ func NewProcessVM(opts *vmOptions) VM {
 }
 
 func (p *ProcessVM) EvaluateAnonymousSnippet(filename string, snippet string) (string, error) {
+	tracer := trace.SpanFromContext(p.ctx).TracerProvider().Tracer("")
+	ctx, span := tracer.Start(p.ctx, "jsonnetsecure.ProcessVM.EvaluateAnonymousSnippet")
```
alnr, line 34:
```suggestion
	ctx, span := tracer.Start(p.ctx, "jsonnetsecure.ProcessVM.EvaluateAnonymousSnippet", trace.WithAttributes(attributesString("filename", filename")))
```

alnr (changes requested):
let's add the filename of the snippet as a tag, immensely helpful for debugging :)

jsonnetsecure/jsonnet_processvm.go
alnr: You get the idea

PR #724, opened 2023-09-13
feat: add geolocation to trace events

alnr:


mszekiel (approved):

zepatrik (commented):
There is even more headers that might come in handy, e.g. `cf-region`. Do we really want to make this a slice? I think that will make it quite hard to use for queries, e.g. query by country. What if (for whatever reason) only the `cf-city` header is set but not `cf-country`? Or the other way around? What if we want to add the region later?

alnr:
> There is even more headers that might come in handy, e.g. `cf-region`. Do we really want to make this a slice? I think that will make it quite hard to use for queries, e.g. query by country. What if (for whatever reason) only the `cf-city` header is set but not `cf-country`? Or the other way around? What if we want to add the region later?

Copypasta from https://github.com/ory/kratos/blob/159c13142a11d4a9cd006da5d0afdda9e0f94ca9/session/session.go#L272-L278

We could factor our this functionality and put it in ory/x. 🤷

zepatrik:
I still think it makes more sense to put city and country in different attributes. For sessions, it is fine to produce a human-readable string instead, as the information is only a hint for users to distinguish sessions. In event attributes however, we want to query by the attribute, which in turn is hard when multiple attributes are mixed into one slice.

alnr:
I don't disagree. Up to @mszekiel. I just wanted to show him where to add this data.

mszekiel:
Okay, I'll take this PR with me as Arne mentioned that's an intro and will continue changing the data structure.

jonas-jonas (approved):

otelx/semconv/context.go
```diff
@@ -36,7 +36,15 @@ func AttributesFromContext(ctx context.Context) []attribute.KeyValue {
 }
 
 func Middleware(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
-	next(rw, r.WithContext(ContextWithAttributes(r.Context(), AttrClientIP(httpx.ClientIP(r)))))
+
```
zepatrik, line 39:
```suggestion
```

zepatrik (approved):

PR #729, opened 2023-10-09
fix: link header in keyset pagination

alnr:
fixes some issues in keyset pagination

### DIFF
--- a/pagination/keysetpagination/header.go
+++ b/pagination/keysetpagination/header.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


jonas-jonas (approved):

pagination/keysetpagination/header.go
```diff
@@ -104,7 +105,7 @@ func Parse(q url.Values, p PageTokenConstructor) ([]Option, error) {
 		}
 		opts = append(opts, WithToken(parsed))
 	}
-	if q.Has("page_size") {
+	if q.Get("page_size") != "" {
```
zepatrik, line 108: Why this? :thinking:

zepatrik (approved):

pagination/keysetpagination/header.go
alnr: I've encountered `GET /admin/identities?page_size=` a bunch of times while working on this. I can also see an SDK generating code like that, and we should be prepared for that. Previously, this would return 400 with an internal error message referencing `strconv.Atoi`. Not it's just equal to default.

PR #730, opened 2023-10-10
chore(popx): improve migration performance by removing unnecessary queries

zepatrik:
This is most relevant during tests, but also Kratos has so many migrations it will significantly speedup it's migration command.

### DIFF
--- a/popx/migrator.go
+++ b/popx/migrator.go
@@ -94,18 +94,24 @@ func (m *Migrator) UpTo(ctx context.Context, step int) (applied int, err error) 
 	defer span.End()
 
 	c := m.Connection.WithContext(ctx)
+	mtn := m.sanitizedMigrationTableName(c)
+
+	var appliedMigrations []migrationRow
+	err = c.RawQuery(fmt.Sprintf("SELECT * FROM %s", mtn)).All(&appliedMigrations)
+	if err != nil && !errIsTableNotFound(err) {
+		return 0, errors.Wrap(err, "migration up: unable to fetch applied migrations")
+	}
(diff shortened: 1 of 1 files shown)


popx/migrator.go
```diff
@@ -94,18 +94,24 @@ func (m *Migrator) UpTo(ctx context.Context, step int) (applied int, err error)
...
 		for _, mi := range mfs {
 			l := m.l.WithField("version", mi.Version).WithField("migration_name", mi.Name).WithField("migration_file", mi.Path)
 
-			exists, err := c.Where("version = ?", mi.Version).Exists(mtn)
-			if err != nil {
-				return errors.Wrapf(err, "problem checking for migration version %s", mi.Version)
-			}
-
-			if exists {
+			if _, exists := appliedMigrationsMap[mi.Version]; exists {
```
aeneasr, line 114: Theoretically we could use an outdated view of already applied migrations, for example a migration that is applied 2 times (for whatever reason) or because there are more than one migration workers. Are we OK accepting that risk for the benefit it brings?

zepatrik: Yes, theoretically this could be an outdated view.
I don't think the benefit will be huge, so if you think this is a possible source of issues it is probably worth the extra few hundred ms.

aeneasr: Sorry, I missed replying to this. I'm not sure to be honest. Migrations should be run as singletons but I guess the question is how do we deal with it if it isn't - so in the case of an operational mistake. Not sure to be honest...

alnr:
I like this PR. Why was it closed? @zepatrik

PR #747, opened 2023-11-13
feat: improved SSRF protection

alnr:


httpx/private_ip_validator.go
hperl, line 1: The function `DisallowIPPrivateAddresses` is now deprecated in favor of the `Control` hook, right? Should we mark it as deprectaed?

hperl (approved):
LGTM! Very nice and clean!

zepatrik (approved):

PR #752, opened 2024-01-04
fix: consider only non-empty values when parsing pagination parameters from URLs

alnr:


alnr:
> Would it make sense to have a page token constructor for UUIDs that rejects invalid UUIDs?

Medium-term, we'll probably end up using the MapPageToken utility more and more. For ordering by creation date and so on. UUID validation makes sense to me in that context 👍

PR #755, opened 2024-01-23
feat: pooled process-isolated Jsonnet VM

alnr:
Go update to 1.21 required me to update snapshots in `jsonschemax`. Looks like to me the previous snapshots were wrong? Weird. Please double check if you understand what's going on.

Adds a `-0` flag to jsonnet subcommand to evaluate multiple snippets from a single process, and a pool with reusable processes.

The design is such that the process pool should be constructed once per program, near `func main()`. Since it's relatively easy to leak processes when the parent process dies, `pool.Close()` is mandatory.

Leaking processes are probably not a big deal in a docker environment, but I might still have an idea on how we can automatically clean them up at least on Linux.

This should drop CPU usage and Jsonnet latency by a huge factor. We might be able to reduce the evaluation timeout (currently 1 sec) down to much lower also.

```
$ go test -bench=. ./jsonnetsecure/
goos: darwin
goarch: arm64
pkg: github.com/ory/x/jsonnetsecure
BenchmarkIsolatedVM-10                       207           7762434 ns/op // current solution: new process per snippet
BenchmarkProcessPoolVM-10                  51032             28746 ns/op // new pooled solution with GOMAXPROCS processes at most
BenchmarkRegularVM-10                      61332             20493 ns/op // vanilla library, no process isolation
BenchmarkReusableProcessVM-10               8208            124614 ns/op // using a single `jsonnet -0` process, this was used for initial benchmarking
PASS
ok      github.com/ory/x/jsonnetsecure  61.678s
```


hperl (approved):
LGTM!

PR #761, opened 2024-02-06
feat: add http client option to disable IPv6

alnr:
* improve connection re-use
* add detailed trace events (DNS, TLS, etc)

Looks like this in Tempo: left is a newly established connection, right is a reused connection.

image

### DIFF
--- a/httpx/ssrf.go
+++ b/httpx/ssrf.go
@@ -4,6 +4,7 @@
 package httpx
 
 import (
+	"context"
 	"net"
 	"net/http"
 	"net/netip"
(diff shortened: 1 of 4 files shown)


hperl (approved):
LGTM!

httpx/ssrf.go
```diff
@@ -20,47 +21,6 @@ type noInternalIPRoundTripper struct {
 	internalIPExceptions        []string
 }
 
-// NewNoInternalIPRoundTripper creates a RoundTripper that disallows
-// non-publicly routable IP addresses, except for URLs matching the given
-// exception globs.
-func NewNoInternalIPRoundTripper(exceptions []string) http.RoundTripper {
```
aeneasr, line 26: Don't remove this if you want to keep BC :)

alnr: Yeah, good point.

PR #762, opened 2024-02-21
fix: don't erroneously stop with nil error when a down migration is found which was never applied

alnr:


popx/migrator.go
```diff
@@ -221,11 +232,11 @@ func (m *Migrator) Down(ctx context.Context, step int) error {
 				legacyVersion := mi.Version[:14]
 				legacyVersionExists, err := c.Where("version = ?", legacyVersion).Exists(mtn)
 				if err != nil {
-					return errors.Wrapf(err, "problem checking for migration version %s", mi.Version)
+					return errors.Wrapf(err, "problem checking for legacy migration version %s", legacyVersion)
 				}
 
 				if !legacyVersionExists {
-					return errors.Wrapf(err, "problem checking for migration version %s", legacyVersion)
+					return errors.Errorf("neither normal (%s) nor legacy migration (%s) exist", mi.Version, legacyVersion)
```
alnr, line 239: Here is the problem. At this point, `err` is always `nil`. `errors.Wrapf` returns nil when its first argument is nil, so this error condition was swallowed and the down migrations silently stopped.

This revelead two incorrect migrations in Kratos. They are uncritical though.

PR #779, opened 2024-04-11
feat: measure external latency

zepatrik:
This adds a middleware and utilities to measure external latency.

### DIFF
--- a/httpx/external_latency.go
+++ b/httpx/external_latency.go
@@ -1,0 +1,33 @@
+// Copyright © 2024 Ory Corp
+// SPDX-License-Identifier: Apache-2.0
+
+package httpx
+
+import (
+	"net/http"
+
+	"github.com/ory/x/reqlog"
+)
+
+// MeasureExternalLatencyTransport is an http.RoundTripper that measures the latency of all requests as external latency.
+type MeasureExternalLatencyTransport struct {
+	Transport http.RoundTripper
+}
+
+var _ http.RoundTripper = (*MeasureExternalLatencyTransport)(nil)
+
+func (m *MeasureExternalLatencyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	upstreamHostPath := req.URL.Scheme + "://" + req.URL.Host + req.URL.Path
+	defer reqlog.StartMeasureExternalCall(req.Context(), "http_request", upstreamHostPath)()
+
+	t := m.Transport
+	if t == nil {
+		t = http.DefaultTransport
+	}
+	return t.RoundTrip(req)
+}
+
+// ClientWithExternalLatencyMiddleware adds a middleware to the client that measures the latency of all requests as external latency.
+func ClientWithExternalLatencyMiddleware(c *http.Client) {
+	c.Transport = &MeasureExternalLatencyTransport{Transport: c.Transport}
+}
--- a/httpx/external_latency.go
+++ b/httpx/external_latency.go
@@ -5,6 +5,7 @@ package httpx
 
 import (
 	"net/http"
+	"time"
 
 	"github.com/ory/x/reqlog"
 )
@@ -18,16 +19,11 @@ var _ http.RoundTripper = (*MeasureExternalLatencyTransport)(nil)
 
 func (m *MeasureExternalLatencyTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	upstreamHostPath := req.URL.Scheme + "://" + req.URL.Host + req.URL.Path
-	defer reqlog.StartMeasureExternalCall(req.Context(), "http_request", upstreamHostPath)()
+	defer reqlog.StartMeasureExternalCall(req.Context(), "http_request", upstreamHostPath, time.Now())
 
 	t := m.Transport
 	if t == nil {
 		t = http.DefaultTransport
 	}
 	return t.RoundTrip(req)
-}
-
-// ClientWithExternalLatencyMiddleware adds a middleware to the client that measures the latency of all requests as external latency.
-func ClientWithExternalLatencyMiddleware(c *http.Client) {
-	c.Transport = &MeasureExternalLatencyTransport{Transport: c.Transport}
 }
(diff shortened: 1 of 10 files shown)


httpx/ssrf.go
```diff
@@ -64,7 +64,7 @@ func init() {
 		ssrf.WithAnyPort(),
 		ssrf.WithNetworks("tcp4", "tcp6"),
 	).Safe
-	prohibitInternalAllowIPv6 = t
+	prohibitInternalAllowIPv6 = &MeasureExternalLatencyTransport{Transport: t}
```
alnr, line 67: Please let's not couple this functionality to the SSRF protection mechanism. We should allow users to opt into the `MeasureExternalLatencyTransport` if they wish. It's as easy as wrapping the transport at the site of `res, err := client.Do(req)`.

alnr (commented):
Does this have to be a middleware? This would work much more simply and be less tightly coupled if we defined a public Context key in the `reqlog` package which gives a `*time.Duration`, to which external calls can add the time they took.

httpx/ssrf.go
zepatrik: This is still opt-in, because the transport is just a no-op when there is no "container" in the context. However, it does make sense to not do this per default for the httpx client.
I was at first thinking about only measuring the external latency on the "prohibitInternal*" variants of this transport, but as it is right now it does not make sense.

zepatrik (commented):
We will always need some kind of middleware to add the data container to the context. As is, the context key is not exposed but there are easy functions to add or get to the data container, I will also add a way to add the data container to a context without the middleware wrapping.

alnr:
> We will always need some kind of middleware to add the data container to the context. As is, the context key is not exposed but there are easy functions to add or get to the data container, I will also add a way to add the data container to a context without the middleware wrapping.

Can we add the context key in the existing reqlog package?

zepatrik:
> Can we add the context key in the existing reqlog package?

What's the benefit of exposing the context key? Or do you mean add a context key for the overall latency as well?

alnr:
Just re-read the code and this already exists in your API essentially 😅

httpx/external_latency.go
```diff
@@ -0,0 +1,33 @@
...
+package httpx
+
+import (
+	"net/http"
+
+	"github.com/ory/x/reqlog"
+)
+
+// MeasureExternalLatencyTransport is an http.RoundTripper that measures the latency of all requests as external latency.
+type MeasureExternalLatencyTransport struct {
```
alnr, line 13: We should get rid of this middleware IMO.

reqlog/external_latency.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"context"
+	"net/http"
+	"sync"
+	"time"
+)
+
+// WithEnableExternalLatencyMeasurement returns a context that measures external latencies.
+func WithEnableExternalLatencyMeasurement(ctx context.Context) context.Context {
```
alnr, line 14: This can be private.

reqlog/external_latency.go
```diff
@@ -0,0 +1,102 @@
...
+	container := contextContainer{
+		latencies: make([]externalLatency, 0),
+	}
+	return context.WithValue(ctx, internalLatencyKey, &container)
+}
+
+// ExternalCallsMiddleware is a middleware that sets up the request context to measure external calls.
+// It has to be used before any other middleware that reads the final external latency.
+func ExternalCallsMiddleware(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
```
alnr, line 23: This does not need an extra middleware, we can just add the context key in the main reqlog middleware.

reqlog/external_latency.go
```diff
@@ -0,0 +1,102 @@
...
+	next(rw, r.WithContext(
+		WithEnableExternalLatencyMeasurement(r.Context()),
+	))
+}
+
+// MeasureExternalCall measures the duration of a function and records it as an external call.
+// The wrapped function's return value is returned.
+func MeasureExternalCall[T any](ctx context.Context, cause, detail string, f func() T) T {
```
alnr, line 31: I feel like we should reduce the API surface and get rid of this and the next wrapper functions.

reqlog/external_latency.go
```diff
@@ -0,0 +1,102 @@
...
+	return func() {
+		container.Lock()
+		defer container.Unlock()
+		container.latencies = append(container.latencies, externalLatency{
+			Took:   time.Since(start),
+			Cause:  cause,
+			Detail: detail,
+		})
+	}
+}
```
alnr, line 61:
```suggestion
func MeasureExternalLatency(ctx context.Context, descriptor string, start time.Time) {
	container, ok := ctx.Value(internalLatencyKey).(*contextContainer)
	if !ok {
		return
	}

	container.Lock()
	defer container.Unlock()
	container.latencies = append(container.latencies, externalLatency{
		Took:   time.Since(start),
		Descriptor: descriptor,
	})
}
```

reqlog/external_latency.go
```diff
@@ -0,0 +1,102 @@
...
+	defer container.Unlock()
+	for _, l := range container.latencies {
+		total += l.Took
+	}
+	return total
+}
+
+// WithDisableExternalLatencyMeasurement returns a context that does not measure external latencies.
+// Use this when you want to disable external latency measurements for a specific request.
+func WithDisableExternalLatencyMeasurement(ctx context.Context) context.Context {
```
alnr, line 83: not needed

reqlog/external_latency.go
```diff
@@ -0,0 +1,102 @@
...
+		container.latencies = append(container.latencies, externalLatency{
+			Took:   time.Since(start),
+			Cause:  cause,
+			Detail: detail,
+		})
+	}
+}
+
+// TotalExternalLatency returns the total duration of all external calls.
+func TotalExternalLatency(ctx context.Context) (total time.Duration) {
```
alnr, line 64: can be private

PR #786, opened 2024-05-22
feat(httpx): use cf-connecting-ip header for client IP

mfzl:
Adds `Cf-Connecting-IP` to the list of headers to check for real client IP. This header is set by Cloudflare to have the real client IP address. 

https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip

Since we are already relying on cloudflare headers for device location and how ubiquitous Cloudflare has become it's a good idea to have support for it.

We cannot rely on `X-forwarded-for` when behind Cloudflare as well, since the order of the IPs listed by Cloudflare is not the order this library checks. 

## Checklist

### DIFF
--- a/httpx/client_info.go
+++ b/httpx/client_info.go
@@ -33,6 +33,8 @@ func GetClientIPAddressesWithoutInternalIPs(ipAddresses []string) (string, error
 func ClientIP(r *http.Request) string {
 	if trueClientIP := r.Header.Get("True-Client-IP"); trueClientIP != "" {
 		return trueClientIP
+	} else if cfConnectingIP := r.Header.Get("Cf-Connecting-IP"); cfConnectingIP != "" {
+		return cfConnectingIP
 	} else if realClientIP := r.Header.Get("X-Real-IP"); realClientIP != "" {
 		return realClientIP
 	} else if forwardedIP := r.Header.Get("X-Forwarded-For"); forwardedIP != "" {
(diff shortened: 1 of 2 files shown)


mfzl:
I could potentially also "fix" the `x-forwarded-for` order as well. It looks like a bug after reading [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For). 

It should be the first public IP address going from left-to-right. Right now it's the first public IP address from right-to-left

aeneasr:
> I could potentially also "fix" the `x-forwarded-for` order as well. It looks like a bug after reading [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For).
> 
> It should be the first public IP address going from left-to-right. Right now it's the first public IP address from right-to-left

Yes please - if there is a bug we appreciate a fix!

aeneasr (approved):

mfzl:
> > I could potentially also "fix" the `x-forwarded-for` order as well. It looks like a bug after reading [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For).
> > It should be the first public IP address going from left-to-right. Right now it's the first public IP address from right-to-left
> 
> Yes please - if there is a bug we appreciate a fix!

Hey @aeneasr I will raise another PR for that. If you have no objections to the changes in this PR this can be merged and closed

aeneasr:
Sounds good!

PR #787, opened 2024-05-27
feat: add WithSpanFromContext logrus helper

jonas-jonas:
## Checklist

### DIFF
--- a/logrusx/helper.go
+++ b/logrusx/helper.go
@@ -95,7 +95,7 @@ func (l *Logger) WithRequest(r *http.Request) *Logger {
 		_, _, spanCtx = otelhttptrace.Extract(r.Context(), r, opts)
 	}
 	if spanCtx.IsValid() {
-		traces := map[string]string{}
+		traces := make(map[string]string, 2)
 		if spanCtx.HasTraceID() {
 			traces["trace_id"] = spanCtx.TraceID().String()
 		}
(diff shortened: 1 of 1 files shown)


logrusx/helper.go
```diff
@@ -107,6 +107,25 @@ func (l *Logger) WithRequest(r *http.Request) *Logger {
 	return ll
 }
 
+func (l *Logger) WithSpan(sc trace.Span) *Logger {
```
zepatrik, line 110: Did you consider passing the full context here instead and then using `trace.SpanFromContext`? I'd guess that that would be the most universal usage.

logrusx/helper.go
```diff
@@ -107,6 +107,25 @@ func (l *Logger) WithRequest(r *http.Request) *Logger {
...
+	if sc == nil {
+		return l
+	}
+	spanCtx := sc.SpanContext()
+	if !spanCtx.IsValid() {
+		return l
+	}
+
+	traces := map[string]string{}
```
zepatrik, line 119:
```suggestion
	traces := make(map[string]string, 2)
```
as we know the max capacity already.

logrusx/helper.go
jonas-jonas: Done.

logrusx/helper.go
jonas-jonas: Done, it was copied from the request helper..

PR #790, opened 2024-06-19
fix: minimum Jsonnet process pool size; bump deps

alnr:
When running with `GOMAXPROX=1` (or on machines with just one CPU), the process pool was previously undersized and would occasionally time out.


aeneasr (commented):
Nice find!

aeneasr (approved):

PR #800, opened 2024-07-30
fix: occasional jsonnet timeouts

alnr:
Reverts #798 and #799 

The fixed 1 second timeout now only applies to the Jsonnet evaluation step,
but not the the spawning of the process which can occasionally take longer.

Additionally, we now warm the pool asynchronously on construction.

### DIFF
--- a/jsonnetsecure/jsonnet.go
+++ b/jsonnetsecure/jsonnet.go
@@ -9,18 +9,11 @@ import (
 	"path"
 	"runtime"
 	"testing"
-	"time"
 
 	"github.com/google/go-jsonnet"
 )
 
 type (
-	evaluationOptions struct {
-		evalTimeout time.Duration
-	}
-
-	EvaluationOptionModifier func(*evaluationOptions)
-
 	VM interface {
 		EvaluateAnonymousSnippet(filename string, snippet string) (json string, formattedErr error)
 		ExtCode(key string, val string)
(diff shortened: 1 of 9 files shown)


jsonnetsecure/jsonnet_pool.go
```diff
@@ -203,6 +208,8 @@ func (vm *processPoolVM) EvaluateAnonymousSnippet(filename string, snippet strin
 		return "", errors.Wrap(err, "jsonnetsecure: acquire")
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
```
zepatrik, line 211: Why should this not be configurable?

alnr: It is: by passing the context to `func (p *DefaultProvider) JsonnetVM(ctx context.Context) (VM, error)` you can choose any timeout you want.

This one-second cooldown is just a hard limit for the evaluation. I've never seen the actual evaluation step taking more than 10ms, except with adversarial workloads.

zepatrik (approved):

jsonnetsecure/jsonnet.go
```diff
@@ -79,12 +70,6 @@ func WithProcessIsolatedVM(ctx context.Context) Option {
 	}
 }
 
-func WithExecTimeout(timeout time.Duration) Option {
```
aeneasr, line 82: If you remove this it needs an update in depending services as well!

alnr: I grepped our code bases and found no uses. :)

PR #810, opened 2024-09-16
feat(otelx): add API key attribute keys

zepatrik:
Adds one attribute key for both project and workspace API keys.


aeneasr (approved):

zepatrik:
Turns out, it is not as easy everywhere to know what type an API key is of.

alnr:
Is this putting actual secrets in attributes?

PR #816, opened 2024-10-07
chore(popx): code simplification and refactoring

zepatrik:
While debugging the migrator, I stumbled upon a couple of improvements that could be made.
I also removed the pkgerx package, as we don't use that now for years.


popx/migrator.go
```diff
@@ -101,44 +101,39 @@ func (m *Migrator) UpTo(ctx context.Context, step int) (applied int, err error)
 		for _, mi := range mfs {
 			l := m.l.WithField("version", mi.Version).WithField("migration_name", mi.Name).WithField("migration_file", mi.Path)
 
-			exists, err := c.Where("version = ?", mi.Version).Exists(mtn)
+			appliedMigrations := make([]string, 0, 2)
+			legacyVersion := mi.Version
+			if len(legacyVersion) > 14 {
+				legacyVersion = legacyVersion[:14]
+			}
+			err := c.RawQuery(fmt.Sprintf("SELECT version FROM %s WHERE version IN (?, ?)", mtn), mi.Version, legacyVersion).All(&appliedMigrations)
```
zepatrik, line 109: We've now combined two queries into one, which should half the number of queries and (depending on the database) can significantly improve the performance of the migrate job.

popx/migrator.go
```diff
@@ -596,13 +586,6 @@ func (m *Migrator) exec(ctx context.Context, fn func() error) error {
 		}
 	}
 
-	if m.Connection.Dialect.Name() == "cockroach" {
-		outer := fn
-		fn = func() error {
-			return crdb.Execute(outer)
```
zepatrik, line 602: From the documentation of `crdb.Execute`:

> It is used to add retry handling to the execution of a single statement. If a multi-statement transaction is being run, use ExecuteTx instead.

But as the transaction is started within, we don't need to retry anything at this point.

popx/migrator_test.go
```diff
@@ -4,144 +4,28 @@
...
 	"github.com/ory/x/logrusx"
-	"github.com/ory/x/pkgerx"
 	. "github.com/ory/x/popx"
-	"github.com/ory/x/sqlcon/dockertest"
 )
 
 //go:embed stub/migrations/transactional/*.sql
 var transactionalMigrations embed.FS
 
-func TestMigratorUpgrading(t *testing.T) {
```
zepatrik, line 29: This test was mostly skipped anyway, and I removed the "legacy" migrator with this PR, so no need to test compatibility anymore.

aeneasr (approved):

popx/migrator.go
aeneasr: With my recent auto-commit changes, we should add retry to migrations that run without transactions (aka go migrations and auto commit transactions).

PR #817, opened 2024-10-18
chore: bump ristretto

zepatrik:
Breaking change: ristretto now uses generics in a breaking way. Any dependents have to also update ristretto and add the generic initializers.


.github/workflows/cve-scan.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
         uses: golang/govulncheck-action@v1
         with:
           go-package: ./...
-          go-version-input: "1.22"
+          go-version-input: "1.23.2"
```
zepatrik, line 33: Also bumped go in the CI because of stdlib vulnerability https://pkg.go.dev/vuln/GO-2024-3106

PR #849, opened 2025-03-26
feat: populate "errors.stack" span attribute when panicking through otelx.End()

alnr:
Added some documentation links too.

### DIFF
--- a/otelx/withspan.go
+++ b/otelx/withspan.go
@@ -8,6 +8,8 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"runtime"
+	"strings"
 
 	pkgerrors "github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


zepatrik (approved):
Why don't we want to strictly follow OpenTelemetry?

alnr:
> Why don't we want to strictly follow OpenTelemetry?

See the comments. OTel suggests emitting a span event on an exception, and, by extension, on panic. However, this is kinda weird IMO. We mark the span as failed via `otelx.End()` when we catch/recover, not when the panic is actually raised.

If we emitted an event on recover, we'd emit events for each frame being unwound, since we typically have `otelx.End()` across the whole call stack. So we'd be spamming a ton of events for each panic.

PR #852, opened 2025-04-25
feat: anonymize JSON data

hperl:
This adds a function `jsonx.Anonymize` to scrub all data from a JSON object while maintaining debuggability.


aeneasr (approved):
very nice! I didn't see a number type - can you add to the test please? Otherwise lgtm

PR #853, opened 2025-04-29
feat(tlsx): add validation for TLS termination

zepatrik:
This PR is moving code from ory/hydra to ory/x:
- https://github.com/ory/hydra/blob/d389fd0269f93c8b7c787f1b3683ae4c6e9e1909/x/tls_termination.go
- https://github.com/ory/hydra/blob/d389fd0269f93c8b7c787f1b3683ae4c6e9e1909/x/tls_termination_test.go

### DIFF
--- a/tlsx/termination.go
+++ b/tlsx/termination.go
@@ -17,43 +17,19 @@ import (
 	"github.com/ory/x/prometheusx"
 )
 
-func MatchesRange(r *http.Request, ranges []string) error {
-	remoteIP, _, err := net.SplitHostPort(r.RemoteAddr)
-	if err != nil {
-		return errors.WithStack(err)
-	}
-
-	xff := strings.Split(r.Header.Get("X-Forwarded-For"), ",")
-	check := make([]string, 1, len(xff)+1)
-	check[0] = remoteIP
-	for _, fwd := range xff {
-		check = append(check, strings.TrimSpace(fwd))
-	}
-
-	for _, rn := range ranges {
-		_, cidr, err := net.ParseCIDR(rn)
-		if err != nil {
-			return errors.WithStack(err)
-		}
-
-		for _, ip := range check {
-			addr := net.ParseIP(ip)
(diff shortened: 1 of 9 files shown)


tlsx/termination.go
```diff
@@ -0,0 +1,91 @@
...
+
+type dependencies interface {
+	logrusx.Provider
+	Writer() herodot.Writer
+}
+
+func EnforceTLSRequests(d dependencies, allowTerminationFrom []string) (negroni.Handler, error) {
+	networks := make([]*net.IPNet, 0, len(allowTerminationFrom))
+	for _, rn := range allowTerminationFrom {
+		_, cidr, err := net.ParseCIDR(rn)
```
hperl, line 28: Nit:

```suggestion
		_, network, err := net.ParseCIDR(rn)
```
`cidr` is the input, not the output.

tlsx/termination.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/ory/x/logrusx"
+	"github.com/ory/x/prometheusx"
+)
+
+type dependencies interface {
+	logrusx.Provider
+	Writer() herodot.Writer
+}
+
+func EnforceTLSRequests(d dependencies, allowTerminationFrom []string) (negroni.Handler, error) {
```
hperl, line 25:
```suggestion
// EnforceTLSRequests creates a middleware that enforces TLS for incoming HTTP requests.
// It allows termination (non-HTTPS traffic) from specific CIDR ranges provided in the `allowTerminationFrom` slice.
// If the request is not secure and does not match the allowed CIDR ranges, an error response is returned.
// The middleware also validates the `X-Forwarded-Proto` header to ensure it is set to "https".
func EnforceTLSRequests(d dependencies, allowTerminationFrom []string) (negroni.Handler, error) {
```

hperl (approved):
LGTM, just one nit and one contribution to the docstring ;)

PR #857, opened 2025-05-13
fix(jonnetsecure): prevent massive memory usage from misbehaving jsonnet script

gaultier:


jsonnetsecure/jsonnet_pool.go
```diff
@@ -129,27 +134,40 @@ func newWorker(ctx context.Context) (_ worker, err error) {
...
 		}
 	}
-	out := make(chan string)
-	go scan(out, stdout)
-	errs := make(chan string)
-	go scan(errs, stderr)
+	out := make(chan string, chanSize)
+	go scan(out, stdoutReader, jsonnetOutputLimit)
+	errs := make(chan string, chanSize)
+	// No limit here because: too much data on stderr is not an error - simply having some already is.
```
aeneasr, line 168: I don't quite understand this, could you elaborate?

jsonnetsecure/jsonnet_processvm.go
```diff
@@ -56,20 +65,51 @@ func (p *ProcessVM) EvaluateAnonymousSnippet(filename string, snippet string) (_
...
 		cmd.Stdin = &stdin
-		cmd.Stdout = &stdout
-		cmd.Stderr = &stderr
 		cmd.Env = []string{"GOMAXPROCS=1"}
 
-		err = cmd.Run()
-		if stderr.Len() > 0 {
-			// If the process wrote to stderr, this means it started and we won't retry.
-			return "", backoff.Permanent(fmt.Errorf("jsonnetsecure: unexpected output on stderr: %q", stderr.String()))
+		stdoutPipe, err := cmd.StdoutPipe()
```
aeneasr, line 70: All errors here now cancel backoff, but then we don't need backoff? Why shouldn't these errors be retried?

aeneasr (changes requested):
Generally nice, but I feel we have lost some retryability in the jsonnet subprocess now. Can you ensure the process is still reliable at scale and potentially retries when we run into e.g. issues because of not enough fd's or whatever?

jsonnetsecure/cmd.go
```diff
@@ -7,18 +7,49 @@ import (
...
 	cmd := &cobra.Command{
 		Use:    "jsonnet",
 		Short:  "Run Jsonnet as a CLI command",
 		Hidden: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// - macos: `setrlimit(2)` with `RLIMIT_AS` seems broken on macOS and its behavior
+			// varies between major versions. Also there is not really a use case for macOS server-side, so we do not
+			// bother.
+			// - windows: This syscall is Unix specific so not available.
+			if !(runtime.GOOS == "windows" || runtime.GOOS == "darwin") {
```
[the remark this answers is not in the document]
gaultier: Some person on FreeBSD or Solaris will surely complain ;) 
If we only support Windows/macOS/Linux officially then yes this would be equivalent.

jsonnetsecure/jsonnet_pool.go
gaultier: Yes maybe it was not super clear but: we limit how much data we read from stdout and stderr to prevent high memory usage. But only the case of `stdout` too big should be considered an error (i.e.: misbehaving script). Stderr most likely comes from the jsonnet vm, not the jsonnet script directly.

jsonnetsecure/jsonnet_processvm.go
gaultier: Good catch! Hmm, I think starting the subprocess (`cmd.Start()`) could be retried because there are not enough file descriptors or such, which is a short-term state. All other errors should be permanent I believe.

gaultier: Good point, will do

jsonnetsecure/cmd.go
```diff
@@ -7,18 +7,49 @@ import (
...
+				limit := syscall.Rlimit{
+					Cur: memoryLimit,
+					Max: memoryLimit,
+				}
+				err := syscall.Setrlimit(syscall.RLIMIT_AS, &limit)
```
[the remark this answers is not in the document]
gaultier: Done 👍

jsonnetsecure/cmd.go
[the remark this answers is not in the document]
gaultier: I split the Windows and non-Windows versions of this function.

jsonnetsecure/jsonnet_processvm.go
gaultier: I extended the tests to run many child processes concurrently including with bad inputs. I believe this should be enough as a first pass.

jsonnetsecure/cmd.go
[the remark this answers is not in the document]
gaultier: The build and tests now pass including on Windows.

zepatrik (approved):

jsonnetsecure/jsonnet_pool.go
```diff
@@ -111,7 +117,7 @@ func newWorker(ctx context.Context) (_ worker, err error) {
 		return worker{}, errors.Wrap(err, "newWorker: failed to create stdin pipe")
 	}
 
-	in := make(chan []byte)
+	in := make(chan []byte, chanSize)
```
alnr, line 120: What is the point of making this channel buffered?

jsonnetsecure/jsonnet_pool.go
```diff
@@ -103,6 +108,7 @@ func newWorker(ctx context.Context) (_ worker, err error) {
 	args, _ := ctx.Value(contextValueArgs).([]string)
 	cmd := exec.Command(path, append(args, "-0")...)
 	cmd.Env = []string{"GOMAXPROCS=1"}
+	cmd.WaitDelay = 100 * time.Millisecond
```
alnr, line 111: Good find!

jsonnetsecure/jsonnet_pool.go
```diff
@@ -129,27 +135,40 @@ func newWorker(ctx context.Context) (_ worker, err error) {
...
 	if err := cmd.Start(); err != nil {
 		return worker{}, errors.Wrap(err, "newWorker: failed to start process")
 	}
 
 	span.SetAttributes(semconv.ProcessPID(cmd.Process.Pid))
 
-	scan := func(c chan<- string, r io.Reader) {
+	scan := func(c chan<- string, r io.Reader, limit int) {
 		defer close(c)
+		// NOTE: `bufio.Scanner` has its own internal limit of 64 KiB.
```
alnr, line 150: Since we split on NUL bytes, the internal buffer limit of bufio.Scanner should be enough to limit the parent's memory usage?

That is, either we read the complete result of the jsonnet evaluation at once if it fits, or we fail with `bufio. ErrBufferFull` without allocating more than the buffer size.

I think we can remove the other read limit.

jsonnetsecure/jsonnet_processvm.go
```diff
@@ -29,6 +35,10 @@ func NewProcessVM(opts *vmOptions) VM {
 	}
 }
 
+// Jsonnet evaluation is run in a subprocess with a timeout, and
+// standard output and error are limited in size.
+// The subprocess is killed when a limit (whichever comes first) is reached.
+// In this case an error is returned.
 func (p *ProcessVM) EvaluateAnonymousSnippet(filename string, snippet string) (_ string, err error) {
```
alnr, line 42: This whole implementation is unused and exists here pretty much for historical reasons only. We only use the process pool implementation since it is orders of magnitude more performant.

I tend to think we should not invest any time in trying to improve this code. Either leave it unchanged or delete outright.

alnr (changes requested):
What happens if we do nothing except 
```
	err := syscall.Setrlimit(syscall.RLIMIT_AS, &lim)
```
in the child process?

Could that be enough?

jsonnetsecure/jsonnet_processvm.go
gaultier: Removed. It was not clear it was fully unused since this is a library.

jsonnetsecure/jsonnet_pool.go
gaultier: If the producer (child process) is faster than the consumer (parent process), memory allocation is unbounded. It is zero-cost to set a limit here to ensure at compile time/review time that this cannot happen.

gaultier:
> What happens if we do nothing except
> 
> ```
> 	err := syscall.Setrlimit(syscall.RLIMIT_AS, &lim)
> ```
> 
> in the child process?
> 
> Could that be enough?

In theory yes, assuming the number of child processes is also bounded (which we do with the pool) *and* that the limit is in effect. 
Unfortunately the latter does not hold - `setrlimit(2)` is not a silver bullet: in this PR it is only called on Unix (not on Windows which has a different mechanism) and macOS has bugs/undocumented behavior around it ([1](https://developer.apple.com/forums/thread/742328), [2](https://developer.apple.com/forums/thread/702803), [3](https://github.com/python/cpython/issues/84698)) so in practice it does not work reliably.  So effectively only Linux (ignoring other minor Unix systems) has a memory limit. 

So... it is frustrating, and having limits everywhere we can as long as the code does not become extremely complex is the only defence in depth we can do IMO.

jsonnetsecure/jsonnet_pool.go
alnr: Better link: https://github.com/ory/x/blob/5c4b146a4829ed039164bcff36c88c556b574f31/jsonnetsecure/jsonnet_pool.go#L181-L195

In other words, if we make this channel buffered, executing not block [here](https://github.com/ory/x/blob/master/jsonnetsecure/jsonnet_pool.go#L184) but instead block [here](https://github.com/ory/x/blob/5c4b146a4829ed039164bcff36c88c556b574f31/jsonnetsecure/jsonnet_pool.go#L191). Is that better? If so, we can make the channel buffer size 1 rather than 16.

alnr:
> In theory yes, assuming the number of child processes is also bounded (which we do with the pool) _and_ that the limit is in effect. Unfortunately the latter does not hold - `setrlimit(2)` is not a silver bullet: in this PR it is only called on Unix (not on Windows which has a different mechanism) and macOS has bugs/undocumented behavior around it ([1](https://developer.apple.com/forums/thread/742328), [2](https://developer.apple.com/forums/thread/702803), [3](https://github.com/python/cpython/issues/84698)) so in practice it does not work reliably. So effectively only Linux (ignoring other minor Unix systems) has a memory limit.
> 

Since we run our infra exclusively on Linux and self-hosted users are responsible for their own Jsonnet snippets, I think limiting on Linux only is good enough.

> So... it is frustrating, and having limits everywhere we can as long as the code does not become extremely complex is the only defence in depth we can do IMO.

`bufio.Scanner` has an internal limit and will return an error if that limit is exceeded. If this happens, we kill the child process anyway.

If we can only limit mem usage in the child reliably on Linux, so be it.

jsonnetsecure/jsonnet_pool.go
```diff
@@ -129,27 +138,40 @@ func newWorker(ctx context.Context) (_ worker, err error) {
...
 		}
 		if err := scanner.Err(); err != nil {
 			c <- "ERROR: scan: " + err.Error()
 		}
 	}
-	out := make(chan string)
-	go scan(out, stdout)
-	errs := make(chan string)
-	go scan(errs, stderr)
+	out := make(chan string, chanSize)
```
alnr, line 169: It would make sense to make `out` and `errs` buffered with size 1, to prevent theoretical goroutine leaks.

jsonnetsecure/jsonnet_pool.go
gaultier: Good idea!

jsonnetsecure/jsonnet_test.go
```diff
@@ -121,39 +149,30 @@ func TestSecureVM(t *testing.T) {
 		})
 	})
 
-	t.Run("case=stack overflow", func(t *testing.T) {
+	t.Run("case=stack overflow pool", func(t *testing.T) {
 		snippet := "local f(x) = if x == 0 then [] else [f(x - 1), f(x - 1)]; f(100)"
-		ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+		_, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
```
alnr, line 154: can be removed

jsonnetsecure/jsonnet.go
```diff
@@ -111,8 +111,15 @@ func (importer *ErrorImporter) Import(importedFrom, importedPath string) (conten
...
+	// We can force the usage of a given jsonnet executable.
+	// Useful to test different versions, or run the tests under wine.
+	if s := os.Getenv("ORY_JSONNET_PATH"); s != "" {
+		return s
+	}
+
 	var stderr bytes.Buffer
-	outPath := path.Join(t.TempDir(), "jsonnet")
+	// Using `t.TempDir()` results in permissions errors on Windows, sometimes.
+	outPath := path.Join(os.TempDir(), "jsonnet")
```
hperl, line 122: `t.TempDir()` is deleted after the test runs, but `os.TempDir()` is not. Do we really need Windows compatibility?

alnr (approved):

aeneasr (approved):

PR #858, opened 2025-05-14
chore: deprecate some code and minor improvements

zepatrik:
All changes should be backwards-compatible.

### DIFF
--- a/fsx/merge.go
+++ b/fsx/merge.go
@@ -224,8 +224,6 @@ func (m *mergedFile) ReadDir(n int) ([]fs.DirEntry, error) {
 		return entries, io.EOF
 	}
 
-	if n <= len(entries) && n > 0 {
-		entries, m.unprocessedDirEntries = entries[:n], entries[n:]
-	}
+	entries, m.unprocessedDirEntries = entries[:n], entries[n:]
 	return entries, nil
 }
--- a/httpx/gzip_server.go
+++ b/httpx/gzip_server.go
@@ -4,7 +4,6 @@
 package httpx
 
 import (
-	"bytes"
 	"compress/gzip"
 	"fmt"
 	"io"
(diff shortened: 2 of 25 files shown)


fsx/merge.go
```diff
@@ -224,8 +224,6 @@ func (m *mergedFile) ReadDir(n int) ([]fs.DirEntry, error) {
 		return entries, io.EOF
 	}
 
-	if n <= len(entries) && n > 0 {
```
zepatrik, line 227: Due to the previous checks, `n > 0 && n < len(entries)` at this point.

httpx/gzip_server.go
```diff
@@ -34,22 +33,13 @@ func (c *CompressionRequestReader) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
 		switch enc = strings.TrimSpace(enc); enc {
 		case "gzip":
-			var b bytes.Buffer
 			reader, err := gzip.NewReader(r.Body)
 			if err != nil {
 				c.ErrHandler(w, r, err)
 				return
 			}
-
-			if _, err := io.Copy(&b, reader); err != nil { //nolint:gosec // FIXME
```
zepatrik, line 44: I guess the fixme comment is related to unlimited body reading, but I don't see why we should read everything here in the first place.

requirex/time_test.go
```diff
@@ -33,15 +33,14 @@ func TestEqualDurationAndTime(t *testing.T) {
 		ok   bool
 		args args
 	}{
-		{ok: false, name: "fail with zero precision", args: args{expected: time.Nanosecond, actual: time.Nanosecond}},
```
zepatrik, line 36: No idea why equal values should fail here, doesn't make sense IMO.

PR #866, opened 2025-06-18
feat(pagination): improved keyset pagination with encrypted tokens

zepatrik:
This re-implements the keysetpagination helpers to support encrypted tokens and includes other improvements, especially around he library's API and capabilities.


zepatrik:
I already have a branch where I use the new version in Keto, and it works great. Migration was very easy.

